PR #2, opened 2020-04-02
Remove extraneous client arg from construction fetcher

yorhodes:
Removes need for API client to be passed to transaction construction fetcher.

### Motivation
API client argument needs to be initialized and configured separately

### Solution
Deletes function argument and uses fetcher instance property

### Open questions
None


heimdall-asguard:
This Pull Request was merged without enough reviews
Please go [here](https://github.com/coinbase/rosetta-sdk-go/issues/3) to review and +1

PR #18, opened 2020-05-01
Generate Code from Rosetta Specifications Repo

### LINKED ISSUE #17: Automatic Code Generation
To update generated code on each release of the Rosetta API, it is necessary to manually download a spec file from the website and include it in the repository. This is a manual and error prone process.   

* Use releases from the recently created [rosetta-specifications](https://github.com/coinbase/rosetta-specifications) repository to generate code instead of using an uploaded specification file.
* Run code generation during build tests and ensure there are no changes. This would occur if a PR updated the version used to generate code without actually generating the code at that version. See rosetta-specifications for an [example](https://github.com/coinbase/rosetta-specifications/blob/master/validate.sh).

None.

patrick-ogrady:
Fixes #17 

### Motivation
* To update generated code on each release of the Rosetta API, it is necessary to manually download a spec file from the website and include it in the repository. This is a manual and error prone process.
* If changes are made to any template files, there is no check to make sure code generation was run again.

### Solution
* Download spec file from the releases on https://github.com/coinbase/rosetta-specifications
* Validate that generated code has been run on spec and template changes before merging


templates/client/api.mustache
```diff
@@ -3,9 +3,10 @@ package {{packageName}}
 
 {{#operations}}
 import (
-	_context "context"
-	_ioutil "io/ioutil"
-	_nethttp "net/http"
+  _context "context"
+  _ioutil "io/ioutil"
+  _nethttp "net/http"
+  "fmt"
```
patrick-ogrady, line 9: Found an issue with a template file when running new check

codegen.sh
```diff
@@ -101,6 +105,9 @@ rm server_tmp/*_service.go
 mv server_tmp/* server;
 rm -rf server_tmp;
 
+# Remove spec file
+rm api.json;
```
niallo, line 109: `rm -f`

niallo: `rm` w/o `-f` can exit non-zero in various ways. Default to `-f` for scripts.

PR #19, opened 2020-05-06
json.RawMessage Metadata

### LINKED ISSUE #16: Replace metadata request map type with json.RawMessage type
It is cumbersome to use the map type for marshalling and unmarshalling within the Celo client and server implementations because validation must be done on an enumerated list of key strings.

https://github.com/coinbase/rosetta-sdk-go/blob/2989e4b865ead7997719ce23c7d1ce3a941d6c74/types/metadata_request.go#L22

I would like to be able to codegen marshallers as defined by the shape of our structs and validate metadata using these marshallers.

None.

**Additional context**
See https://github.com/celo-org/rosetta/blob/c51cfc731a13672a34f78cafff42978bea3574c9/service/rpc/servicer.go#L397

patrick-ogrady:
Fixes #16 .

### Motivation
To use the metadata fields extensively, it would be very useful to be able to encode/decode the field using the `encoding/json` library (or any other JSON encoding library). This would prevent the need to step through each key of the `map[string]interface{}` to populate a custom struct.

### Solution
This PR changes all `map[string]interface{}` usage to `json.RawMessage`. Because `json.RawMessage` is not necessarily a JSON object (as required by 1.3.1 of the specification), this PR also adds asserts all `json.RawMessage` fields are JSON objects.

### Other Improvements
* Add a deterministic `Hash` function for comparing struct equality when containing `json.RawMessage`. `reflect.DeepEqual` does not always work if a struct contains `json.RawMessage`.
* Return client configuration as `Configuration` instead of a list of items.
* Add/Subtract `Amount.Value` utility function

### Future PR
Find a cleaner way to create a deterministic hash of any struct in `types`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,7 @@ deps: | gen-deps
 gen-deps:
 	${GO_INSTALL} github.com/google/addlicense
 	${GO_INSTALL} github.com/segmentio/golines
+	${GO_INSTALL} golang.org/x/tools/cmd/goimports
 
 gen:
 	./codegen.sh
@@ -30,9 +31,11 @@ lint: | lint-examples
 
 format:
 	gofmt -s -w -l .
+	goimports -w .
 
 check-format:
 	! gofmt -s -l . | read
+	! goimports -l . | read
 
 test:
 	${TEST_SCRIPT}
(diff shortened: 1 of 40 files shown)


heimdall-asguard:
Review Error for annieke @ 2020-05-06 17:36:40 UTC
User failed mfa authentication, see [go/mfa-help](http://go/mfa-help)

PR #24, opened 2020-05-12
Sort OperationGroup Results

patrick-ogrady:
### Motivation
It can be very useful for clients to receive a **deterministic ordering** of `OperationGroups` and `Operations` within an `OperationGroup` when calling `GroupOperations`.

### Solution
Sort all `OperationGroups` in a transaction by the lowest `OperationIdentifier.Index` in each group. Then, sort all `Operations` in an `OperationGroup` by `OperationIdentifier.Index`, ascending.

### DIFF
--- a/parser/group_operations.go
+++ b/parser/group_operations.go
@@ -41,6 +41,7 @@ func containsInt(valid []int, value int) bool {
 	return false
 }
 
+// addOperationToGroup appends a *types.Operation to an *OperationGroup.
 func addOperationToGroup(
 	destination *OperationGroup,
 	destinationIndex int,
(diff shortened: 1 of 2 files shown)


parser/group_operations.go
```diff
@@ -66,11 +69,43 @@ func addOperationToGroup(
 	}
 }
 
+// sortOperationGroups returns a slice of OperationGroups sorted by the lowest
+// OperationIdentifier.Index in each group. This function also sorts all
+// operations in each OperationGroup by OperationIdentifier.Index. It can be
+// useful to consumers to have a deterministic ordering of groups and ops within
+// each group.
+func sortOperationGroups(opLen int, opGroups map[int]*OperationGroup) []*OperationGroup {
```
annieke, line 77: Adding a note here on why `map[int]` instead of a slice would be nice

PR #30, opened 2020-05-18
Return Correct Error on Context Cancellation

patrick-ogrady:
### Motivation
* If context is cancelled, retry methods in the `fetcher` package return an incorrect error (exhausted retries instead of context error).
* If context is cancelled, the reconciler can silence the `ctx.Err()` and return `nil` instead.

### Solution
Return the correct error.


patrick-ogrady:
Decrease in coverage is deceptive because tests were added to areas of the code that were previously untested.

PR #54, opened 2020-06-26
Add Keys package

juliankoh:
### Motivation
Added a `keys` package to rosetta-sdk-go, which includes keygen, verify, and sign methods for Secp256k1 and Edwards25519 curves. This will be primarily used within tests for implementations of the Construction API, where we can sign & verify the constructed transactions.


patrick-ogrady:
As we discussed, rebasing on #55 will remove the need for redefining all the `CurveType` and `SignatureType` in this package.

keys/signing.go
```diff
@@ -0,0 +1,147 @@
...
+			return nil, fmt.Errorf("sign: unable to sign. %w", err)
+		}
+		sig = sig[:64]
+	}
+	return sig, nil
+}
+
+// signEd25519 is a helper function to sign payloads with the Ed25519 SignatureType
+func signEd25519(privKeyBytes, payload []byte) []byte {
+	privKey := ed25519.NewKeyFromSeed(privKeyBytes)
```
jakecraige, line 50: [From docs](https://golang.org/pkg/crypto/ed25519/#NewKeyFromSeed), this will panic if the private key is the wrong size, do we want to run the `KeyPair.isValid()` at some point before calling this to ensure we have a properly formatted key before using it here?

keys/types.go
```diff
@@ -0,0 +1,29 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package keys
+
+import "github.com/coinbase/rosetta-sdk-go/types"
+
+// PrivateKey contains the hex-encoded privkey bytes as well as the CurveType
+type PrivateKey struct {
+	HexBytes  string
```
jakecraige, line 21: Why are we storing strings in the struct instead of bytes in the SDK? It seems preferable to store them in bytes sine Go supports it and only do the hex conversion at the API layer.

It would also help prevent all this extra decoding you're having to do to use the libraries 😏

keys/signing_test.go
```diff
@@ -0,0 +1,109 @@
...
+		HexBytes:      "68656C6C6F0D0A",
+		SignatureType: signatureType,
+	}
+
+	signature, err := SignPayload(payload, keypair)
+	assert.NoError(t, err)
+
+	verify, err := Verify(signature)
+	assert.NoError(t, err)
+	assert.Equal(t, verify, true)
```
jakecraige, line 48: This doesn't really matter, but there is also an `assert.True` that you could use here if you want 😂

keys/signing_test.go
```diff
@@ -0,0 +1,109 @@
...
+	assert.NoError(t, err)
+
+	signatureType := types.Ed25519
+	payload := &types.SigningPayload{
+		Address:       "test",
+		HexBytes:      "68656C6C6F0D0A",
+		SignatureType: signatureType,
+	}
+
+	signature, err := SignPayload(payload, keypair)
```
jakecraige, line 43: Do you think it's worth making any assertions on the signature length since that's one of the things that changese between sig types?

keys/signing.go
```diff
@@ -0,0 +1,147 @@
...
+	"encoding/hex"
+	"fmt"
+
+	"github.com/coinbase/rosetta-sdk-go/asserter"
+	"github.com/coinbase/rosetta-sdk-go/types"
+	"github.com/ethereum/go-ethereum/crypto/secp256k1"
+)
+
+// signECDSA is a helper function to sign payloads with the ECDSA SignatureType
+func signECDSA(privKeyBytes, payload []byte, signatureType types.SignatureType) ([]byte, error) {
```
itstehkman, line 28: nit: these function definitions are pretty procedural with tons of switch cases. Thoughts on making them more object-oriented? I think it would be cleaner, since it would get rid of all the switch cases and clean up the function definitions to follow a more uniform interface.

We could have a `Key` class with a curve attribute (`Secp256k1 ` or `Edwards25519 `), which each have valid signature types to be used with it.

Then the Key class could have Sign functions that take a signatureType, since a curve can be used with a few different signature types.

This follows the model that golang sets in `ed25519` - https://golang.org/pkg/crypto/ed25519/#PrivateKey.Sign

keys/keys.go
```diff
@@ -0,0 +1,110 @@
...
+
+	default:
+		return nil, fmt.Errorf("%s is not supported", curve)
+	}
+
+	return keyPair, nil
+}
+
+// IsValid checks the validity of a keypair
+func (k KeyPair) IsValid() (bool, error) {
```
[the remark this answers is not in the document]
patrick-ogrady: I agree with @jakecraige's suggestion. We follow the `return error or nil if it is valid` approach for everything in the `asserter` package. 

https://github.com/coinbase/rosetta-sdk-go/blob/b9ec88dfefda0124e4ee41079b81ef2f0f6d7004/asserter/block.go#L35-L60

keys/keys_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+func TestKeypairValidity(t *testing.T) {
+	// Non matching curves
+	keyPair, _ := GenerateKeypair(types.Edwards25519)
+	keyPair.PublicKey.CurveType = types.Secp256k1
+	valid, err := keyPair.IsValid()
+	assert.Equal(t, false, valid)
+	assert.Contains(t, err.Error(), "do not match")
+
+	// Privkey length too short
```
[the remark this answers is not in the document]
patrick-ogrady: Highly recommend table driven tests!

https://github.com/coinbase/rosetta-sdk-go/blob/b9ec88dfefda0124e4ee41079b81ef2f0f6d7004/asserter/block_test.go#L27-L65

keys/signing.go
juliankoh: I think for now we have a small enough number of curves/signatures where a switch seems manageable. When we do decide to add more, I'll definitely consider this refactor!

keys/keys_test.go
```diff
@@ -0,0 +1,73 @@
...
+	errMsg  string
+}
+
+func mockKeyPair(privKey []byte, curveType types.CurveType) *KeyPair {
+	keypair, _ := GenerateKeypair(curveType)
+	keypair.PrivateKey.Bytes = privKey
+	return keypair
+}
+
+var privKeyTests = []privKeyTest{
```
patrick-ogrady, line 53: I would move this into `TestKeypairValidity`

keys/signing.go
patrick-ogrady: I agree with @itstehkman here @juliankoh. Although it is manageable now with this small set of signature schemes, it will make it **much** easier for external contributors to add new schemes in the future. I was thinking something simple like this:
```
func GenerateKeypair(CurveType) Signer

interface Signer {
  PublicKey()
  Sign(payload, SignatureType) -> secp256k1 has multiple signature types
  Verify(signature)
}
```

Suggested File Structure (file name based on curve):
```
signer.go (where interface is defined)
edwards25519.go
secp256k1.go (implementation for both ECDSA and ECDSA pubkey recovery lives here)
```

keys/keys.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	// Checks if pk and sk have the same CurveType
+	if pkCurve != skCurve {
+		return fmt.Errorf("private key curve %s and public key curve %s do not match", skCurve, pkCurve)
+	}
+
+	// All privkeys are 32-bytes
+	if len(sk) != btcec.PrivKeyBytesLen {
+		return fmt.Errorf("invalid privkey length %v", len(sk))
```
itstehkman, line 106: small nit: should we add the expected length to the error message? Can do in a later PR.

keys/signer_ed25519.go
```diff
@@ -0,0 +1,82 @@
...
+func (s SignerEd25519) Verify(signature *types.Signature) (bool, error) {
+	pubKey, err := hex.DecodeString(signature.PublicKey.HexBytes)
+	if err != nil {
+		return false, fmt.Errorf("verify: unable to decode pubkey. %w", err)
+	}
+	decodedMessage, err := hex.DecodeString(signature.SigningPayload.HexBytes)
+	if err != nil {
+		return false, fmt.Errorf("verify: unable to decode message. %w", err)
+	}
+	decodedSignature, err := hex.DecodeString(signature.HexBytes)
```
jakecraige, line 70: Is it possible to also update `types.Signature`, `types.SigningPayload` and `types.PublicKey` type to use `[]byte` instead so you can avoid all this decoding?

Since these already exist I'd expect this to potentially have larger implications and we could follow up with it, but I think the amount of extra decoding necessary shows that the way we're storing this data doesn't match how it's used, so making that closer could be beneficial.

keys/signer_ed25519_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+var signerEd25519 Signer
+
+func init() {
+	keypair, _ := GenerateKeypair(types.Edwards25519)
+	signerEd25519 = Signer(SignerEd25519{*keypair})
+}
+
+func TestSignEd25519(t *testing.T) {
+	signatureType := types.EcdsaRecovery
```
jakecraige, line 33: This is not the signature type we want here 😏 , seems like we need some sort of extra validation if this isn't error-ing since an Ed25519 signer can't produce this type of signature

keys/signer_ed25519_test.go
```diff
@@ -0,0 +1,49 @@
...
+	payload := types.SigningPayload{
+		Address:       "test",
+		HexBytes:      "68656C6C6F0D0A",
+		SignatureType: signatureType,
+	}
+
+	signature, err := signerEd25519.Sign(payload)
+	assert.NoError(t, err)
+	signatureBytes, _ := hex.DecodeString(signature.HexBytes)
+	assert.Equal(t, len(signatureBytes), 64)
```
jakecraige, line 44: There is also an `assert.Len` if you want to use the fancy helper functions

keys/signer_ed25519.go
```diff
@@ -0,0 +1,82 @@
...
+	KeyPair KeyPair
+}
+
+// PublicKey returns the PublicKey of the signer
+func (s SignerEd25519) PublicKey() types.PublicKey {
+	return *s.KeyPair.PublicKey
+}
+
+// Signs arbitrary payloads using a KeyPair
+func (s SignerEd25519) Sign(payload types.SigningPayload) (*types.Signature, error) {
```
jakecraige, line 37: The way we're using pointers or not is inconsistent, in some cases we return pointers and others not, and I can't seem to identify why we're using one vs the other. What's the decision criteria going into this?

In my go experience, typically people just use struct pointers basically everywhere 🤷

juliankoh: Basically I tried to use non-pointers as much as I could, which I looked at https://golang.org/pkg/crypto/#Signer and https://pkg.go.dev/golang.org/x/crypto/ed25519?tab=doc for inspiration. Is that correct? My previous version used pointers everywhere, but don't really know what Go convention is here

keys/signer_ed25519.go
juliankoh: @patrick-ogrady thoughts here?

keys/signer_ed25519.go
patrick-ogrady: I’ve read a ton of articles advocating for each approach (pointers vs non-pointers).

For consistency with the rest of rosetta-sdk-go, I believe we should use pointers (except for CurveType and SignatureType) for both function inputs and returned values @juliankoh

keys/signer_ed25519.go
patrick-ogrady: If each signer implementation takes `[]byte` instead of these Rosetta types, you would only need to implement the decoding logic once in some wrapper (which has the Rosetta types as args) around any signer implementation. This approach could lead to less redundant encoding/decoding code and make the signer implementations more minimal (probably preferred).

Still thinking about the best format for the wrapper but a naive approach could be something as simple as taking a signer as the first argument in a collection of functions with Rosetta types (the same functions on the Signer interface).

How do you feel about this @juliankoh?

keys/signer.go
```diff
@@ -0,0 +1,24 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package keys
+
+import "github.com/coinbase/rosetta-sdk-go/types"
+
+// Signer is an interface for different curve signers
+type Signer interface {
```
patrick-ogrady, line 20: Related to https://github.com/coinbase/rosetta-sdk-go/pull/54/files#r449375144 ... let's use pointers for `types.PublicKey` and `types.SigningPayload`

keys/signer_ed25519.go
```diff
@@ -0,0 +1,82 @@
...
+	"encoding/hex"
+	"fmt"
+
+	"github.com/coinbase/rosetta-sdk-go/asserter"
+	"github.com/coinbase/rosetta-sdk-go/types"
+)
+
+// SignerEd25519 is initialized from a keypair
+type SignerEd25519 struct {
+	KeyPair KeyPair
```
patrick-ogrady, line 28: nit: let's use a pointer here

keys/signer_ed25519.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// PublicKey returns the PublicKey of the signer
+func (s SignerEd25519) PublicKey() types.PublicKey {
+	return *s.KeyPair.PublicKey
```
patrick-ogrady, line 33: With a pointer we won't need too do this de-reference (I try to avoid de-references as much as possible)

keys/signer_ed25519.go
```diff
@@ -0,0 +1,82 @@
...
+		return nil, err
+	}
+	privKeyBytes := s.KeyPair.PrivateKey.Bytes
+	decodedMessage, err := hex.DecodeString(payload.HexBytes)
+	if err != nil {
+		return nil, fmt.Errorf("sign: unable to decode message. %w", err)
+	}
+
+	privKey := ed25519.NewKeyFromSeed(privKeyBytes)
+	sig := ed25519.Sign(privKey, decodedMessage)
```
patrick-ogrady, line 49: I'd add an assertion here that the `SigningPayload.SignatureType` is `empty` or that it is `ed25519` : https://github.com/coinbase/rosetta-sdk-go/pull/54/files#r449318027

keys/signer_secp256k1_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+var signerSecp256k1 Signer
+
+func init() {
+	keypair, _ := GenerateKeypair(types.Secp256k1)
+	signerSecp256k1 = Signer(SignerSecp256k1{*keypair})
+}
+
+func TestSignSecp256k1EcdsaRecovery(t *testing.T) {
+	signatureType := types.EcdsaRecovery
```
patrick-ogrady, line 39: I'd also test that `types.Ed25519` does not work.

patrick-ogrady: Should be pretty easy to do if you convert to a table-driven test.

keys/types.go
```diff
@@ -0,0 +1,29 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package keys
+
+import "github.com/coinbase/rosetta-sdk-go/types"
+
+// PrivateKey contains the hex-encoded privkey bytes as well as the CurveType
+type PrivateKey struct {
+	Bytes     []byte
```
patrick-ogrady, line 21: @juliankoh thoughts on making this `hex` so that we could store as JSON. If we use bytes here, we can't store as easily.

jakecraige: @patrick-ogrady bytes serializes to JSON just fine in Go, except it in base64, so the hex standardization choice makes things more difficult 😛

keys/signer_ed25519.go
jakecraige: > This approach could lead to less redundant encoding/decoding code and make the signer implementations more minimal (probably preferred).

jakecraige: I don't understand enough about the SDK structure to comment much on using a wrapper type or not, but I'd strongly recommend against passing around hex strings as much as possible. 

Each one is an opportunity to have a bug in decoding logic and makes getting a correct implementation more difficult. When we're decoding things like private keys, it's even more important to make sure we don't make mistakes, so requiring extra effort to decode is a risk. If we use the strong types to our advantage and use `[]byte` instead in every place _except the API layer_ where we can't because of JSON, we end up with a safer system that's easier to integrate with and less likely to have bugs.

patrick-ogrady: I resolved this issue in #60. `types.Signature`, `types.SigningPayload`, and `types.PublicKey` all now use `[]byte`! Thanks for the feedback @jakecraige.

keys/types.go
patrick-ogrady: I should've dug more deeply into this topic before writing this comment @jakecraige, my bad.

@juliankoh let's change this to use `[]byte` again but use the custom marshaler I wrote (in #60) to serialize as hex instead of base64 to be consistent with other byte representations in the SDK.

Here is an example of the custom marshaling logic I use: https://github.com/coinbase/rosetta-sdk-go/blob/cd4a060d72c34ce5391198e65f3bdb42890fa7dd/types/public_key.go#L31-L70

keys/keys.go
```diff
@@ -0,0 +1,109 @@
...
+// limitations under the License.
+
+package keys
+
+import (
+	"crypto/ed25519"
+	"fmt"
+
+	"github.com/btcsuite/btcd/btcec"
+	"github.com/coinbase/rosetta-sdk-go/asserter"
```
patrick-ogrady, line 22: nit...you should move `github.com/coinbase/rosetta-sdk-go/asserter` and `github.com/coinbase/rosetta-sdk-go/types` to their own section in import ahead of any external imports.

```
import (
	"crypto/ed25519"
	"fmt"

	"github.com/coinbase/rosetta-sdk-go/asserter"
        "github.com/coinbase/rosetta-sdk-go/types"

	"github.com/btcsuite/btcd/btcec"
)
```

keys/keys.go
```diff
@@ -0,0 +1,109 @@
...
+		return err
+	}
+
+	// Checks if pk and sk have the same CurveType
+	if pkCurve != skCurve {
+		return fmt.Errorf("private key curve %s and public key curve %s do not match", skCurve, pkCurve)
+	}
+
+	// All privkeys are 32-bytes
+	if len(sk) != btcec.PrivKeyBytesLen {
```
patrick-ogrady, line 105: I'd suggest making your own constant for this `PrivKeyBytesLen`. Using `btcec`'s constant is confusing because we don't generate `ed25519` keys in it.

I'd also add a comment stating that this could change as more `CurveType`s are added.

keys/signer_ed25519.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/coinbase/rosetta-sdk-go/types"
+)
+
+// SignerEd25519 is initialized from a keypair
+type SignerEd25519 struct {
+	KeyPair *KeyPair
+}
+
+// PublicKey returns the PublicKey of the signer
+func (s SignerEd25519) PublicKey() *types.PublicKey {
```
patrick-ogrady, line 31: I'd recommend doing `(s *SignerEd25519)` here...you should do the same below for `secp256k1`

keys/signer_ed25519.go
```diff
@@ -0,0 +1,83 @@
...
+
+// Signs arbitrary payloads using a KeyPair
+func (s SignerEd25519) Sign(payload *types.SigningPayload, sigType types.SignatureType) (*types.Signature, error) {
+	err := s.KeyPair.IsValid()
+	if err != nil {
+		return nil, err
+	}
+
+	if !(payload.SignatureType == types.Ed25519 || payload.SignatureType == "") {
+		return nil, fmt.Errorf("sign: payload signature type is not ed25519")
```
patrick-ogrady, line 43: nit: you may want to use the constant in the error message instead of typing out `ed25519`...same can be done on 47

keys/signer_ed25519.go
```diff
@@ -0,0 +1,83 @@
...
+	message := signature.SigningPayload.Bytes
+	sig := signature.Bytes
+	err := asserter.Signatures([]*types.Signature{signature})
+	if err != nil {
+		return err
+	}
+
+	verify := ed25519.Verify(pubKey, message, sig)
+	if !verify {
+		return fmt.Errorf("verify: verify returned false")
```
patrick-ogrady, line 79: you should use `errors.New` if you are not expanding any variables in the error string

keys/signer_secp256k1.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/ethereum/go-ethereum/crypto/secp256k1"
+)
+
+// SignerSecp256k1 is initialized from a keypair
+type SignerSecp256k1 struct {
+	KeyPair *KeyPair
+}
+
+// PublicKey returns the PublicKey of the signer
+func (s SignerSecp256k1) PublicKey() *types.PublicKey {
```
patrick-ogrady, line 32: nit: see comment above about `(s *SignerSecp256k1)`

keys/signer_secp256k1.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// Signs arbitrary payloads using a KeyPair
+func (s SignerSecp256k1) Sign(payload *types.SigningPayload, sigType types.SignatureType) (*types.Signature, error) {
+	err := s.KeyPair.IsValid()
+	if err != nil {
+		return nil, err
+	}
+	privKeyBytes := s.KeyPair.PrivateKey.Bytes
+
```
patrick-ogrady, line 43: nit: you should do a similar check here for ensuring `payload.SignatureType == sigType`... I'd recommend making a helper function as this will be pretty common

keys/types.go
patrick-ogrady, line 21: Just to confirm, we are ok serializing private key bytes as base64? @juliankoh

keys/types.go
```diff
@@ -0,0 +1,29 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package keys
+
+import "github.com/coinbase/rosetta-sdk-go/types"
+
+// PrivateKey contains the hex-encoded privkey bytes as well as the CurveType
+type PrivateKey struct {
```
patrick-ogrady, line 20:
```suggestion
type PrivateKey struct { // I'd recommend populating the json key names in case the variable name changes
	Bytes     []byte    `json:"hex_bytes"`
	CurveType CurveType `json:"curve_type"`
}

// MarshalJSON overrides the default JSON marshaler
// and encodes bytes as hex instead of base64.
func (pk *PrivateKey) MarshalJSON() ([]byte, error) {
	type Alias PrivateKey
	j, err := json.Marshal(struct {
		Bytes string `json:"hex_bytes"`
		*Alias
	}{
		Bytes: hex.EncodeToString(pk.Bytes),
		Alias: (*Alias)(pk),
	})
	if err != nil {
		return nil, err
	}
	return j, nil
}

// UnmarshalJSON overrides the default JSON unmarshaler
// and decodes bytes from hex instead of base64.
func (pk *PrivateKey) UnmarshalJSON(b []byte) error {
	type Alias PrivateKey
	r := struct {
		Bytes string `json:"hex_bytes"`
		*Alias
	}{
		Alias: (*Alias)(pk),
	}
	err := json.Unmarshal(b, &r)
	if err != nil {
		return err
	}

	bytes, err := hex.DecodeString(r.Bytes)
	if err != nil {
		return err
	}

	pk.Bytes = bytes
	return nil
}
```

keys/types.go
patrick-ogrady: Resolved with code suggestion

keys/types.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"encoding/hex"
+	"encoding/json"
+
+	"github.com/coinbase/rosetta-sdk-go/types"
+)
+
+// KeyPair contains a PrivateKey and its' associated PublicKey
+type KeyPair struct {
+	PublicKey  *types.PublicKey
```
patrick-ogrady, line 26: nit: I'd add `json` keys here too:
```
type KeyPair struct {
	PublicKey  *types.PublicKey `json:"public_key"`
	PrivateKey *PrivateKey `json:"private_key"`
}
```

keys/keys_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+func TestGenerateKeypairEd25519(t *testing.T) {
+	curve := types.Edwards25519
+	keypair, err := GenerateKeypair(curve)
+
+	assert.NoError(t, err)
+	assert.Equal(t, keypair.PublicKey.CurveType, curve)
+	assert.Equal(t, keypair.PrivateKey.CurveType, curve)
+
+	privKeyJSON, _ := json.Marshal(keypair.PrivateKey)
```
patrick-ogrady, line 49: nit: I'd `assert.NoError` on this marshal

keys/keys_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+func TestGenerateKeypairSecp256k1(t *testing.T) {
+	curve := types.Secp256k1
+	keypair, err := GenerateKeypair(curve)
+
+	assert.NoError(t, err)
+	assert.Equal(t, keypair.PublicKey.CurveType, curve)
+	assert.Equal(t, keypair.PrivateKey.CurveType, curve)
+
+	privKeyJSON, _ := json.Marshal(keypair.PrivateKey)
```
patrick-ogrady, line 34: nit: I'd `assert.NoError` on this marshal

keys/signer_secp256k1.go
```diff
@@ -0,0 +1,104 @@
...
+// Signs arbitrary payloads using a KeyPair
+func (s *SignerSecp256k1) Sign(payload *types.SigningPayload, sigType types.SignatureType) (*types.Signature, error) {
+	err := s.KeyPair.IsValid()
+	if err != nil {
+		return nil, err
+	}
+	privKeyBytes := s.KeyPair.PrivateKey.Bytes
+
+	if payload.SignatureType != sigType {
```
patrick-ogrady, line 45: nit: this will error if a `sigType` is passed in when `payload.SignatureType` is not populated. You should use the same logic as here: https://github.com/coinbase/rosetta-sdk-go/pull/54/files#diff-e27a53bd15dfa9595612a921778dd566R43-R45

keys/keys_test.go
```diff
@@ -0,0 +1,87 @@
...
+	var privKey PrivateKey
```
patrick-ogrady, line 35: just to make sure `hex` marshaling is working, I'd add this test for both the `secp256k1` and `edwards25519` keypair:
```suggestion
        // Simple Hex Check
	simpleType := struct {
		HexBytes string `json:"hex_bytes"`
	}{}

	err = json.Unmarshal(j, &simpleType)
	assert.NoError(t, err)

	b, err := hex.DecodeString(simpleType.HexBytes)
	assert.NoError(t, err)

	assert.Equal(t, s.Bytes, b)
```

keys/keys_test.go
```diff
@@ -0,0 +1,87 @@
...
+	err = json.Unmarshal(privKeyJSON, &privKey)
+	assert.NoError(t, err)
+	assert.Equal(t, privKey.Bytes, keypair.PrivateKey.Bytes)
```
patrick-ogrady, line 38: nit: expected should go first:
```suggestion
	assert.Equal(t, keypair.PrivateKey.Bytes, privKey.Bytes)
```

keys/keys_test.go
```diff
@@ -0,0 +1,87 @@
...
```
patrick-ogrady, line 53: nit: expected should go first:
```suggestion
	assert.Equal(t, keypair.PrivateKey.Bytes, privKey.Bytes)
```

keys/keys.go
```diff
@@ -0,0 +1,113 @@
...
+	"crypto/ed25519"
+	"fmt"
+
+	"github.com/coinbase/rosetta-sdk-go/asserter"
+	"github.com/coinbase/rosetta-sdk-go/types"
+
+	"github.com/btcsuite/btcd/btcec"
+)
+
+// Currently all suppported privkeys are 32-bytes
```
patrick-ogrady, line 27: nit: comment should start with `PrivKeyBytesLen`

patrick-ogrady:
Make sure to add the `keys` package here so all tests run correctly: https://github.com/coinbase/rosetta-sdk-go/blob/cd4a060d72c34ce5391198e65f3bdb42890fa7dd/Makefile#L14-L15

You should also add a line item to the `README`

keys/keys_test.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func TestGenerateKeypairEd25519(t *testing.T) {
```
patrick-ogrady, line 41: nit: this should be `TestGenerateKeypairEdwards25519`

keys/keys_test.go
```diff
@@ -0,0 +1,98 @@
...
+	secpKeypair, _ := GenerateKeypair(types.Secp256k1)
+	ed25519Keypair, _ := GenerateKeypair(types.Secp256k1)
+
+	var keyPairs = []*KeyPair{secpKeypair, ed25519Keypair}
+	for _, keypair := range keyPairs {
+		privKeyJSON, err := json.Marshal(keypair.PrivateKey)
+		assert.NoError(t, err)
+
+		// Simple Hex Check
+		simpleType := struct {
```
patrick-ogrady, line 37: ❤️

keys/signer_secp256k1.go
```diff
@@ -0,0 +1,106 @@
...
+		sig, err = secp256k1.Sign(payload.Bytes, privKeyBytes)
+		if err != nil {
+			return nil, fmt.Errorf("sign: unable to sign. %w", err)
+		}
+	case types.Ecdsa:
+		sig, err = secp256k1.Sign(payload.Bytes, privKeyBytes)
+		if err != nil {
+			return nil, fmt.Errorf("sign: unable to sign. %w", err)
+		}
+		sig = sig[:64]
```
patrick-ogrady, line 66: nit: use a constant for all references of `64`

PR #66, opened 2020-07-14
Inactive Reconciler Performance when --lookup-balance-by-block=false

### LINKED ISSUE #65: rosetta-cli check with --lookup-balance-block=false doesn't find reconciliation errors unless --interesting-accounts is given
**Describe the bug**

In testing of a Rosetta implementation on a blockchain that should contain reconciliation errors, the check command fails to find the errors unless --interesting-accounts is used. This seems to be caused by the reconciliation checks becoming overly delayed so as to never happen.

**To Reproduce**

Run `rosetta-cli check --lookup-balance-by-block=false --interesting-accounts interesting.json` and watch it find reconciliation errors.

Run `rosetta-cli check --lookup-balance-by-block=false` and let it spin forever and never find any reconciliation problems.

**Expected behavior**

I expect the same errors to be found, though perhaps in different amounts of time.

patrick-ogrady:
### Motivation
Fix #65 

### Solution
* Ensure all seen accounts are added to the `inactiveAccountQueue` even if we are far behind tip
* Ensure accounts aren't leaked from the `inactiveAccountQueue` (could occur if accounts
* Improve performance by changing frequent operation (`seenAccounts` check on each block) from `O(n)` to `~O(1)` (caused reconciler to get much slower as more accounts seen).
* Ensure `inactive reconciler` respects the `highWaterMark` (so useless queries are not performed during syncing).
* Fix edge case issue where an account could be added to the queue multiple times (in different go routines). This fix was possible because a lookup is performed while holding the lock and that lookup is now `~O(1)` time now.

### Other Thoughts
We could further optimize the `reconciler` to cache balances it retrieved when behind tip so that reconciliation could begin before making it to tip. This would require some careful thinking and testing and should be attempted in another PR.

### DIFF
--- a/reconciler/configuration.go
+++ b/reconciler/configuration.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 
 	"github.com/coinbase/rosetta-sdk-go/parser"
+	"github.com/coinbase/rosetta-sdk-go/types"
 )
 
 // Option is used to overwrite default values in
@@ -59,7 +60,7 @@ func WithSeenAccounts(seen []*AccountCurrency) Option {
 			r.inactiveQueue = append(r.inactiveQueue, &InactiveEntry{
 				Entry: acct,
 			})
-			r.seenAccounts = append(r.seenAccounts, acct)
+			r.seenAccounts[types.Hash(acct)] = struct{}{}
 		}
 
 		fmt.Printf(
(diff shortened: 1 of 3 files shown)


heimdall-asguard:
Review Error for juliankoh @ 2020-07-15 11:03:36 UTC
User failed mfa authentication, public email is not set on your github profile. see [go/mfa-help](http://go/mfa-help)

heimdall-asguard:
Review Error for juliankoh @ 2020-07-15 11:04:24 UTC
User failed mfa authentication, public email is not set on your github profile. see [go/mfa-help](http://go/mfa-help)

PR #69, opened 2020-07-16
Fix a race condition on measuring length of inactiveQueue

slowboat0:
Fixed a race condition where the size of the `inactiveQueue` could have been mutated by the time we log it


patrick-ogrady:
Nice find @slowboat0! Fortunately, this bug would only lead to logging inaccuracy.

PR #71, opened 2020-07-16
Fix a race condition that can lead to logging inaccuracy

slowboat0:
Fixed a race condition where the size of the `inactiveQueue` could have been mutated by the time it's logged
which leads to logging inaccuracy


patrick-ogrady:
Nice find @slowboat0! Fortunately, this bug could only lead to logging inaccuracy.

PR #72, opened 2020-07-16
Bug when asserting ConstructionParseResponses

### LINKED ISSUE #73: Bug in Asserter.ConstructionParseResponse
**Describe the bug**

Asserter.ConstructionParseResponse asserts that `len(response.Signers) == 0` for all `ConstructionParseResponse`. Instead, we need to raise an error when `len(response.Signers) == 0` if the tx is signed, and `len(response.Signers) > 0` if the tx is unsigned. 

**To Reproduce**

**Expected behavior**

**Additional context**

juliankoh:
Fixes #73 

### Summary
Fixes a bug where `Asserter.ConstructionParseResponse` raises an error if there are no signers, when it shouldn't for unsigned transactions.

### DIFF
--- a/asserter/construction.go
+++ b/asserter/construction.go
@@ -111,8 +111,8 @@ func ConstructionHashResponse(
 // not have a valid set of operations or
 // if the signers is empty.
 func (a *Asserter) ConstructionParseResponse(
-	signed bool,
 	response *types.ConstructionParseResponse,
+	signed bool,
 ) error {
 	if a == nil {
 		return ErrAsserterNotInitialized
--- a/asserter/construction.go
+++ b/asserter/construction.go
@@ -111,6 +111,7 @@ func ConstructionHashResponse(
 // not have a valid set of operations or
 // if the signers is empty.
 func (a *Asserter) ConstructionParseResponse(
+	signed bool,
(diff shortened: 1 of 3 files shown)


asserter/construction.go
```diff
@@ -111,6 +111,7 @@ func ConstructionHashResponse(
 // not have a valid set of operations or
 // if the signers is empty.
 func (a *Asserter) ConstructionParseResponse(
+	signed bool,
```
patrick-ogrady, line 114: nit: can we do `response, signed` to keep consistency with all other Response validation functions?

asserter/construction_test.go
```diff
@@ -299,55 +299,132 @@ func TestConstructionParseResponse(t *testing.T) {
...
+			if test.err != nil {
+				assert.Error(t, err)
+				assert.Contains(t, err.Error(), test.err.Error())
+			} else {
+				assert.NoError(t, err)
+			}
+		})
+	}
+
+	var unsignedTest = map[string]struct {
```
patrick-ogrady, line 360: nit: you should add another value to the struct in `tests` (`signed bool`) instead of duplicating the entire testing loop here.

PR #76, opened 2020-07-20
[CHORE] Update comparison match logic for opposite amounts

qiwu7:
Fixes # .

### Motivation
Update comparison match logic for opposite amounts
- Add more validation on match operations
- Improve time complexity of comparison logic from O(N^2) -> O(N)

### Solution
use a single for loop instead of nested to simply things

### DIFF
--- a/parser/match_operations.go
+++ b/parser/match_operations.go
@@ -301,8 +301,8 @@ func operationMatch(
 // equalAmounts returns an error if a slice of operations do not have
 // equal amounts.
 func equalAmounts(ops []*types.Operation) error {
-	if len(ops) <= 1 {
-		return fmt.Errorf("cannot check equality of %d operations", len(ops))
+	if len(ops) == 0 {
+		return errors.New("cannot check equality of 0 operations")
 	}
 
 	val, err := types.AmountValue(ops[0].Amount)
@@ -437,49 +437,35 @@ func comparisonMatch(
 		if err := matchIndexValid(matches, amountMatch[1]); err != nil {
 			return fmt.Errorf("%w: opposite amounts comparison error", err)
 		}
-		if err := matchOperationsForOppositeComparisonValid(amountMatch[0], matches); err != nil {
-			return err
+
+		match0Ops := matches[amountMatch[0]].Operations
+		match1Ops := matches[amountMatch[1]].Operations
+		if err := equalAmounts(match0Ops); err != nil {
+			return fmt.Errorf(
+				"%w: opposite amounts comparison error for match index %d",
+				err,
+				amountMatch[0],
+			)
 		}
-		if err := matchOperationsForOppositeComparisonValid(amountMatch[1], matches); err != nil {
-			return err
+		if err := equalAmounts(match1Ops); err != nil {
+			return fmt.Errorf(
+				"%w: opposite amounts comparison error for match index %d",
+				err,
+				amountMatch[1],
+			)
 		}
 
 		// only need to check opposites amount for the very first operation from each
 		// matched operations group since we made sure all amounts within the same
 		// matched operation group are the same
 		if err := oppositeAmounts(
-			matches[amountMatch[0]].Operations[0],
-			matches[amountMatch[1]].Operations[0],
+			match0Ops[0],
+			match1Ops[0],
 		); err != nil {
 			return fmt.Errorf("%w: amounts not opposites", err)
 		}
 	}
 
-	return nil
-}
-
-// matchOperationsForOppositeComparisonValid checks if the matched operations have at least one
-// element and
-// checks if all the operations have the same amount
-func matchOperationsForOppositeComparisonValid(matchIndex int, matches []*Match) error {
-	matchOps := matches[matchIndex].Operations
-
-	if len(matchOps) == 0 {
-		return fmt.Errorf(
-			"opposite amounts comparison error: no matched operations for match index %d",
-			matchIndex,
-		)
-	}
-	// check if all operations within the same match slice have the same amount
-	if len(matchOps) > 1 {
-		if err := equalAmounts(matchOps); err != nil {
-			return fmt.Errorf(
-				"%w: opposite amounts comparison error: amounts not equal within index %d match operations",
-				err,
-				matchIndex,
-			)
-		}
-	}
 	return nil
 }
 
--- a/parser/match_operations.go
+++ b/parser/match_operations.go
@@ -437,15 +437,40 @@ func comparisonMatch(
 		if err := matchIndexValid(matches, amountMatch[1]); err != nil {
 			return fmt.Errorf("%w: opposite amounts comparison error", err)
 		}
-		for _, op := range matches[amountMatch[0]].Operations {
-			for _, otherOp := range matches[amountMatch[1]].Operations {
-				if err := oppositeAmounts(
-					op,
-					otherOp,
-				); err != nil {
-					return fmt.Errorf("%w: amounts not opposites", err)
-				}
+
+		match1Ops := matches[amountMatch[0]].Operations
+		match2Ops := matches[amountMatch[1]].Operations
+
+		if len(match1Ops) != len(match2Ops) {
+			return errors.New("opposite amounts comparison error: unequal length between match amounts")
+		}
+		// continue if nothing to compare
+		if len(match1Ops) == 0 {
+			continue
+		}
+		// check if all operations within the same match slice have the same amount
+		for i := 1; i < len(match1Ops); i++ {
+			if err := equalAmounts([]*types.Operation{
+				match1Ops[i-1],
+				match1Ops[i],
+			}); err != nil {
+				return fmt.Errorf(
+					"%w: opposite amounts comparison error: amounts not equal within the same match amounts array", err)
 			}
+			if err := equalAmounts([]*types.Operation{
+				match2Ops[i-1],
+				match2Ops[i],
+			}); err != nil {
+				return fmt.Errorf(
+					"%w: opposite amounts comparison error: amounts not equal within the same match amounts array", err)
+			}
+		}
+		// check for opposites amount
+		if err := oppositeAmounts(
+			match1Ops[0],
+			match2Ops[0],
+		); err != nil {
+			return fmt.Errorf("%w: amounts not opposites", err)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


parser/match_operations.go
```diff
@@ -437,16 +437,41 @@ func comparisonMatch(
...
-					op,
-					otherOp,
-				); err != nil {
-					return fmt.Errorf("%w: amounts not opposites", err)
-				}
+
+		match1Ops := matches[amountMatch[0]].Operations
+		match2Ops := matches[amountMatch[1]].Operations
+
+		if len(match1Ops) != len(match2Ops) {
```
patrick-ogrady, line 444: nit: there is no expectation that `match1Ops` and `match2Ops` must be equal length

parser/match_operations.go
```diff
@@ -437,16 +437,41 @@ func comparisonMatch(
...
+			return errors.New("opposite amounts comparison error: unequal length between match amounts")
+		}
+		// continue if nothing to compare
+		if len(match1Ops) == 0 {
+			continue
+		}
+		// check if all operations within the same match slice have the same amount
+		for i := 1; i < len(match1Ops); i++ {
```
patrick-ogrady, line 452:
```suggestion
		for i := 0; i < len(match1Ops) - 1; i++ {
		  if err := equalAmounts([]*types.Operation{
				match1Ops[i],
				match1Ops[i+1],
			}); err != nil {
				return fmt.Errorf(
					"%w: opposite amounts comparison error: amounts not equal within the same match amounts array", err)
			}
```

patrick-ogrady, line 452:
```suggestion
		equalAmounts(match1Ops)
		
```

parser/match_operations.go
```diff
@@ -437,16 +437,41 @@ func comparisonMatch(
...
 			}
 		}
+		// check for opposites amount
+		if err := oppositeAmounts(
+			match1Ops[0],
+			match2Ops[0],
+		); err != nil {
+			return fmt.Errorf("%w: amounts not opposites", err)
+		}
 	}
```
patrick-ogrady, line 476: Code would be something like:
```
equalAmounts(match1Ops)
equalAmounts(match2Ops)

oppositeAmounts(match1Ops[0], match2Ops[0])
```

parser/match_operations.go
qiwu7: Interesting. It does not make sense to me from a caller's perspective that there are different num of Pos & Neg operations with in a single Transaction. 

Or maybe this method is not restricted to the use case of a transaction? It just doing the comparison of the all amounts and make sure they are all opposite amounts, and it's up to the caller to determine what to do when they are unequal length matches in match result.

I think it makes sense if the above is true.

parser/match_operations.go
```diff
@@ -437,15 +437,28 @@ func comparisonMatch(
...
+		match1Ops := matches[amountMatch[0]].Operations
+		match2Ops := matches[amountMatch[1]].Operations
+
+		if len(match1Ops) == 0 || len(match2Ops) == 0 {
+			return errors.New("opposite amounts comparison error: no matched operations")
+		}
+		// check if all operations within the same match slice have the same amount
+		if err := equalAmounts(match1Ops); err != nil {
+			return fmt.Errorf(
+				"%w: opposite amounts comparison error: amounts not equal within the same match amounts array", err)
```
patrick-ogrady, line 450: nit: for these errors can you specify not equal within the `nth` description

parser/match_operations.go
```diff
@@ -437,15 +437,28 @@ func comparisonMatch(
...
-			for _, otherOp := range matches[amountMatch[1]].Operations {
-				if err := oppositeAmounts(
-					op,
-					otherOp,
-				); err != nil {
-					return fmt.Errorf("%w: amounts not opposites", err)
-				}
-			}
+
+		match1Ops := matches[amountMatch[0]].Operations
```
patrick-ogrady, line 441: can we 0-index this to be consistent with variable creation

parser/match_operations.go
```diff
@@ -437,16 +437,33 @@ func comparisonMatch(
...
+		match1Ops := matches[amountMatch[1]].Operations
+
+		if len(match0Ops) == 0 || len(match1Ops) == 0 {
+			return errors.New("opposite amounts comparison error: no matched operations")
+		}
+		// check if all operations within the same match slice have the same amount
+		if len(match0Ops) > 1 {
+			if err := equalAmounts(match0Ops); err != nil {
+				return fmt.Errorf(
+					"%w: opposite amounts comparison error: amounts not equal within the 0th match operations", err)
```
patrick-ogrady, line 451: nit: I would use %d here and insert the value `amountMatch[0]` (not guaranteed to be the `0th`

parser/match_operations.go
```diff
@@ -437,16 +437,33 @@ func comparisonMatch(
...
-					otherOp,
-				); err != nil {
-					return fmt.Errorf("%w: amounts not opposites", err)
-				}
+
+		match0Ops := matches[amountMatch[0]].Operations
+		match1Ops := matches[amountMatch[1]].Operations
+
+		if len(match0Ops) == 0 || len(match1Ops) == 0 {
+			return errors.New("opposite amounts comparison error: no matched operations")
```
patrick-ogrady, line 445: On further thought, I like the idea of breaking this into 2 if statements so we can surface a clearer error message.

nit: description `amountMatch[0]` does not have enough operations

parser/match_operations.go
```diff
@@ -437,21 +437,48 @@ func comparisonMatch(
...
+	matchOps := matches[matchIndex].Operations
+
+	if len(matchOps) == 0 {
+		return fmt.Errorf(
+			"opposite amounts comparison error: no matched operations for match index %d",
+			matchIndex,
+		)
+	}
+	// check if all operations within the same match slice have the same amount
+	if len(matchOps) > 1 {
```
patrick-ogrady, line 470: nit: you can remove this if statement as we assume it must be true (assuming `matchIndex` is not negative, which it should never be)

parser/match_operations.go
```diff
@@ -437,21 +437,48 @@ func comparisonMatch(
...
-		for _, op := range matches[amountMatch[0]].Operations {
-			for _, otherOp := range matches[amountMatch[1]].Operations {
-				if err := oppositeAmounts(
-					op,
-					otherOp,
-				); err != nil {
-					return fmt.Errorf("%w: amounts not opposites", err)
-				}
-			}
+		if err := validateMatchOperationsForOppositeComparison(amountMatch[0], matches); err != nil {
```
patrick-ogrady, line 440: nit: can we switch the order to be consistent with other functions here (like `matchIndexValid`)

parser/match_operations.go
```diff
@@ -437,21 +437,48 @@ func comparisonMatch(
...
+			return err
+		}
+		if err := validateMatchOperationsForOppositeComparison(amountMatch[1], matches); err != nil {
+			return err
+		}
+		// check for opposites amount
```
patrick-ogrady, line 446: nit[STYLE]: should be enter above this line

nit: can we write a clearer comment here?

patrick-ogrady:
@qiwu7 to fix the `check-gen` error you need to run `make shorten-lines` I think!

parser/match_operations.go
qiwu7: equalAmounts method will return err if length of the operations is less than 1.
https://github.com/coinbase/rosetta-sdk-go/blob/master/parser/match_operations.go#L304-L306

patrick-ogrady: We already check if `len(matchOps) < 1` on line 467 in your PR.

parser/match_operations.go
```diff
@@ -437,21 +437,52 @@ func comparisonMatch(
...
-		for _, op := range matches[amountMatch[0]].Operations {
-			for _, otherOp := range matches[amountMatch[1]].Operations {
-				if err := oppositeAmounts(
-					op,
-					otherOp,
-				); err != nil {
-					return fmt.Errorf("%w: amounts not opposites", err)
-				}
-			}
+		if err := matchOperationsForOppositeComparisonValid(amountMatch[0], matches); err != nil {
```
patrick-ogrady, line 440: On further thought, I suggest we remove `matchOperationsForOppositeComparisonValid` because we are just wrapping `equalAmounts` with the conditionals it already contains (which I think is a bad practice).

```suggestion
		if err := equalAmounts(matches[amountMatch[0]].Operations); err != nil {
		  return fmt.Errorf("%w: amounts not equal within index %d match operations", err, amountMatch[0])
		}
```

patrick-ogrady: If we want to surface better errors in these cases (length 0 vs length 1), I'd suggest modifying `equalAmounts` itself: https://github.com/coinbase/rosetta-sdk-go/blob/f929bfc4a262afbbb4053570664be00053a0b657/parser/match_operations.go#L304-L306

patrick-ogrady: If we still want to allow `len(ops) == 1` to be considered equal in some cases, we could add another arg

patrick-ogrady: Still thinking if we'd want to consider operation amounts `equal` if there is only 1 op.

heimdall-asguard:
Review Error for patrick-ogrady @ 2020-07-22 20:05:58 UTC
User failed mfa authentication, see [go/mfa-help](http://go/mfa-help)

parser/match_operations.go
```diff
@@ -301,8 +301,8 @@ func operationMatch(
 // equalAmounts returns an error if a slice of operations do not have
 // equal amounts.
 func equalAmounts(ops []*types.Operation) error {
-	if len(ops) <= 1 {
-		return fmt.Errorf("cannot check equality of %d operations", len(ops))
+	if len(ops) == 0 {
```
patrick-ogrady, line 304: I would've expected this test to trigger a test change...can we write a simple one for testing equal amounts with 1 op? @qiwu7

qiwu7: will do

qiwu7: actually there already are test cases for `equalAmounts` with only 1 op, that's why the test was failing previously.

patrick-ogrady: I have to put up a PR that touches this code, so I'm going to approve this and I'll add the test there! This PR has been back and forth enough already haha.

qiwu7: cool. I can help review that when it's ready :)

PR #83, opened 2020-07-25
Assert Successful Intent

patrick-ogrady:
### Changes
* Modify `ExpectedOperations` to allow for asserting success on matched operations
* Print out all missing matches in `ExpectedOperations` error

### DIFF
--- a/parser/intent.go
+++ b/parser/intent.go
@@ -15,6 +15,7 @@
 package parser
 
 import (
+	"errors"
 	"fmt"
 	"log"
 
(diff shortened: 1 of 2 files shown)


parser/intent.go
```diff
@@ -73,11 +77,24 @@ func ExpectedOperations(
 				continue
 			}
 
-			if ExpectedOperation(in, obs) == nil {
-				matches[i] = struct{}{}
-				foundMatch = true
-				break
+			err := ExpectedOperation(in, obs)
+			if err != nil {
+				continue
```
slowboat0, line 82: Do we not care about this error? Looks like it's not being logged or returned anywhere

patrick-ogrady: @slowboat0 just updated code with a comment around this to explain!

PR #89, opened 2020-08-06
Optimize Syncer

patrick-ogrady:
Related Issue: https://github.com/coinbase/rosetta-cli/issues/87

This PR greatly improves sync performance by changing syncing to be continuous instead of discrete (syncing in 1000 block chunks). Most performance gains come from keeping our IO resources in high use (previously, we wouldn't attempt to store anything while fetching the block range). I will post a chart describing the magnitude of sync speedup when merging this into the `rosetta-cli`.

As a side effect, we greatly reduce memory usage by avoiding caching up to 1000 blocks in memory before processing (**when the default concurrency is used, this is a memory usage reduction of 99%**).

### Changes
- [x] Make syncer continuous instead of discrete
- [x] Improve syncer test coverage
- [x] Change `BlockConcurrency` to `SyncConcurrency`
- [x] Remove `BlockConcurrency` from `Fetcher`
- [x] Cleanup logic in syncer and in tests


syncer/syncer.go
```diff
@@ -231,38 +247,158 @@ func (s *Syncer) processBlock(
 	return nil
 }
 
-func (s *Syncer) syncRange(
+// addIndicies appends a range of indicies (from
```
qiwu7, line 250:
```suggestion
// addBlockIndicies appends a range of indicies (from
```

syncer/syncer.go
```diff
@@ -231,38 +247,158 @@ func (s *Syncer) processBlock(
...
 		}
+
+		if s.nextIndex < lastProcessed && reorgStart == -1 {
+			reorgStart = lastProcessed
+		}
+	}
+
+	return nil
+}
+
```
qiwu7, line 365: How about adding a comment here? something like
```suggestion
syncRange fetches and processes a range of blocks concurrently starting from the next index to the given endIndex inclusively 
```

syncer/syncer.go
```diff
@@ -231,38 +247,158 @@ func (s *Syncer) processBlock(
...
+	go func() {
+		_ = g.Wait()
+		close(results)
+	}()
+
+	cache := make(map[int64]*types.Block)
+	for b := range results {
+		cache[b.BlockIdentifier.Index] = b
+
+		if err := s.processBlocks(ctx, cache, endIndex); err != nil {
```
qiwu7, line 394: just curious why do we need to iterate over the results while `s.processBlocks` is also processing from s.nextIndex to endIndex?

patrick-ogrady: We are pulling from an unordered channel here. Here is an example:
```
s.nextIndex = 7
result -> block 10
result -> block 8
result -> block 9
result -> block 7 
-> process block 7-10
```

patrick-ogrady: We want to process all retrieved but unprocessed blocks each time we take something out of the channel.

patrick-ogrady: Definitely the most complex part of this implementation (which is why I wrote so many tests).

PR #91, opened 2020-08-08
Support Omitted Blocks

patrick-ogrady:
Blocked by: https://github.com/coinbase/rosetta-specifications/pull/35 and `v1.4.2` release

This PR adds support for syncing omitted blocks and for populating the new suggested fee-related fields that were added in `v1.4.2`.

### Changes
- [x] Update to `v1.4.2`
- [x] Allow nil block in fetcher
- [x] Assert suggested fee, max fee, suggested fee multiplier
- [x] Modify syncer to support index gaps

### DIFF
--- a/asserter/account.go
+++ b/asserter/account.go
@@ -42,9 +42,9 @@ func ContainsCurrency(currencies []*types.Currency, currency *types.Currency) bo
 func assertUniqueAmounts(amounts []*types.Amount) error {
 	currencies := make([]*types.Currency, 0)
 	for _, amount := range amounts {
-		// Ensure a currency is used at most once in balance.Amounts
+		// Ensure a currency is used at most once
 		if ContainsCurrency(currencies, amount.Currency) {
-			return fmt.Errorf("currency %+v used in balance multiple times", amount.Currency)
+			return fmt.Errorf("currency %+v used multiple times", amount.Currency)
 		}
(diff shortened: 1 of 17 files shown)


asserter/account.go
```diff
@@ -35,17 +35,16 @@ func ContainsCurrency(currencies []*types.Currency, currency *types.Currency) bo
...
+func assertUniqueAmounts(amounts []*types.Amount) error {
 	currencies := make([]*types.Currency, 0)
 	for _, amount := range amounts {
-		// Ensure a currency is used at most once in balance.Amounts
+		// Ensure a currency is used at most once
 		if ContainsCurrency(currencies, amount.Currency) {
-			return fmt.Errorf("currency %+v used in balance multiple times", amount.Currency)
+			return fmt.Errorf("currency %+v used multiple times", amount.Currency)
 		}
 		currencies = append(currencies, amount.Currency)
```
slowboat0, line 49: The complexity of `assertUniqueAmounts` is `O(n^2)`. We should be able to do `O(n)`

patrick-ogrady: That's a good call out. This is some old code I haven't touched in a while.

patrick-ogrady: Usually this array is at most 1 or 2 items but could potentially be much larger when implementations support multiple tokens.

We can punt on this until then!

PR #93, opened 2020-08-11
Add ImportPrivKey function to keys package

juliankoh:
### Motivation
This adds the `ImportPrivKey` function to the keys package, and allows users to pass in a hex-encoded privkey string to get a KeyPair.

This will be used in `rosetta-cli` where users can specify their prefunded addresses' privkeys in hex format in their config files.

### DIFF
--- a/keys/keys.go
+++ b/keys/keys.go
@@ -16,6 +16,7 @@ package keys
 
 import (
 	"crypto/ed25519"
+	"encoding/hex"
 	"fmt"
 
 	"github.com/coinbase/rosetta-sdk-go/asserter"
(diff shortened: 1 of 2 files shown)


keys/keys.go
```diff
@@ -27,6 +28,49 @@ import (
 // PrivKeyBytesLen are 32-bytes for all supported curvetypes
 const PrivKeyBytesLen = 32
 
+// ImportPrivKey returns a Keypair from a hex-encoded privkey string
+func ImportPrivKey(privKeyHex string, curve types.CurveType) (*KeyPair, error) {
```
patrick-ogrady, line 32: I'd recommend using `keys.PrivateKey` here because it already has curve and does the hex decoding for you!

patrick-ogrady: JK not possible, we don't have a `keys.PrivateKey` type.

PR #98, opened 2020-08-12
exposing Rosetta types.Error in fetcher

cindyxkuang:
Fixes # .

### Motivation

In order to allow consumers of the fetcher to parse errors upstream, we need to expose the Rosetta types.Errors being returned from our API calls.

### Solution

Defined an ErrorRes struct that wraps both the Go error and Rosetta error being returned from API calls, and returning that from the fetcher instead.

### DIFF
--- a/fetcher/fetcher.go
+++ b/fetcher/fetcher.go
@@ -80,6 +80,13 @@ type Fetcher struct {
 	retryElapsedTime       time.Duration
 }
 
+// Error wraps the two possible types of error responses returned
+// by the Rosetta Client
+type Error struct {
+	Err       error        `json:"err"`
+	ClientErr *types.Error `json:"client_err"`
+}
+
 // New constructs a new Fetcher with provided options.
 func New(
 	serverAddress string,
--- a/fetcher/fetcher.go
+++ b/fetcher/fetcher.go
@@ -175,7 +175,7 @@ func (f *Fetcher) InitializeAsserter(
 		networkOptions,
 	)
 	if assertErr != nil {
-		return nil, nil, err
+		return nil, nil, &Error{Err: assertErr}
 	}
 	f.Asserter = newAsserter
 
--- a/fetcher/fetcher.go
+++ b/fetcher/fetcher.go
@@ -132,10 +132,10 @@ func (f *Fetcher) InitializeAsserter(
 ) (
 	*types.NetworkIdentifier,
 	*types.NetworkStatusResponse,
-	error,
+	*Error,
 ) {
 	if f.Asserter != nil {
-		return nil, nil, errors.New("asserter already initialized")
+		return nil, nil, &Error{Err: errors.New("asserter already initialized")}
 	}
 
 	// Attempt to fetch network list
@@ -145,7 +145,7 @@ func (f *Fetcher) InitializeAsserter(
 	}
 
 	if len(networkList.NetworkIdentifiers) == 0 {
-		return nil, nil, ErrNoNetworks
+		return nil, nil, &Error{Err: ErrNoNetworks}
 	}
 	primaryNetwork := networkList.NetworkIdentifiers[0]
 
@@ -169,14 +169,15 @@ func (f *Fetcher) InitializeAsserter(
 		return nil, nil, err
 	}
 
-	f.Asserter, err = asserter.NewClientWithResponses(
+	newAsserter, assertErr := asserter.NewClientWithResponses(
 		primaryNetwork,
 		networkStatus,
 		networkOptions,
 	)
-	if err != nil {
+	if assertErr != nil {
 		return nil, nil, err
 	}
+	f.Asserter = newAsserter
 
 	return primaryNetwork, networkStatus, nil
 }
--- a/fetcher/mempool.go
+++ b/fetcher/mempool.go
@@ -28,20 +28,27 @@ import (
 func (f *Fetcher) Mempool(
 	ctx context.Context,
 	network *types.NetworkIdentifier,
-) ([]*types.TransactionIdentifier, error) {
-	response, _, err := f.rosettaClient.MempoolAPI.Mempool(
+) ([]*types.TransactionIdentifier, *Error) {
+	response, clientErr, err := f.rosettaClient.MempoolAPI.Mempool(
 		ctx,
 		&types.NetworkRequest{
 			NetworkIdentifier: network,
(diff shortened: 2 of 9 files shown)


fetcher/network.go
```diff
@@ -24,26 +24,39 @@ import (
 	"github.com/coinbase/rosetta-sdk-go/types"
 )
 
+type ErrorRes struct {
```
patrick-ogrady, line 27: I recommend changing this to `type Error` and changing the field names to the ones below:

```suggestion
type Error struct {
	error           error
	clientError *types.Error
}
```

patrick-ogrady: I think these fields should be public (i.e. capitalized).

fetcher/network.go
```diff
@@ -62,16 +75,20 @@ func (f *Fetcher) NetworkStatusRetry(
 	)
 
 	for {
-		networkStatus, err := f.NetworkStatus(
+		networkStatus, errRes := f.NetworkStatus(
```
patrick-ogrady, line 78: nit: I'd keep this as `err`

fetcher/network.go
```diff
@@ -67,22 +80,22 @@ func (f *Fetcher) NetworkStatusRetry(
...
 		)
-		if errors.Is(err, ErrAssertionFailed) {
-			return nil, fmt.Errorf("%w: /network/status not attempting retry", err)
-		}
-
 		if err == nil {
 			return networkStatus, nil
 		}
 
+		if errors.Is(err.Err, ErrAssertionFailed) {
```
patrick-ogrady, line 87: ❤️

fetcher/network.go
```diff
@@ -96,23 +109,32 @@ func (f *Fetcher) NetworkStatusRetry(
 }
 
 // NetworkList returns the validated response
-// from the NetworList method.
+// from the NetworkList method.
```
patrick-ogrady, line 112: ❤️

fetcher/network.go
```diff
@@ -158,28 +180,37 @@ func (f *Fetcher) NetworkListRetry(
...
-// from the NetworList method.
+// from the NetworkOptions method.
 func (f *Fetcher) NetworkOptions(
 	ctx context.Context,
 	network *types.NetworkIdentifier,
 	metadata map[string]interface{},
-) (*types.NetworkOptionsResponse, error) {
-	NetworkOptions, _, err := f.rosettaClient.NetworkAPI.NetworkOptions(
+) (*types.NetworkOptionsResponse, *Error) {
+	networkOptions, clientErr, err := f.rosettaClient.NetworkAPI.NetworkOptions(
```
patrick-ogrady, line 189: ❤️

patrick-ogrady:
This is **GREAT** work @cindyxkuang! Looking forward to seeing this applied to the other endpoints.

fetcher/network.go
```diff
@@ -24,26 +24,39 @@ import (
 	"github.com/coinbase/rosetta-sdk-go/types"
 )
 
+type Error struct {
```
patrick-ogrady, line 27: This should be in `fetcher/fetcher.go` or `fetcher/types.go`

fetcher/network.go
```diff
@@ -158,28 +180,37 @@ func (f *Fetcher) NetworkListRetry(
 }
 
 // NetworkOptions returns the validated response
-// from the NetworList method.
+// from the NetworkOptions method.
```
qiwu7, line 183: 🤟

patrick-ogrady: 🤦  clearly too much copy and paste in the early days

patrick-ogrady:
Can we return a `*fetcher.Error` in the `*Retry` methods too?

fetcher/fetcher.go
```diff
@@ -162,14 +169,15 @@ func (f *Fetcher) InitializeAsserter(
...
 
-	f.Asserter, err = asserter.NewClientWithResponses(
+	newAsserter, assertErr := asserter.NewClientWithResponses(
 		primaryNetwork,
 		networkStatus,
 		networkOptions,
 	)
-	if err != nil {
+	if assertErr != nil {
 		return nil, nil, err
```
patrick-ogrady, line 178: nit: should be `assertErr`

fetcher/mempool.go
```diff
@@ -28,20 +28,27 @@ import (
...
+) ([]*types.TransactionIdentifier, *Error) {
+	response, clientErr, err := f.rosettaClient.MempoolAPI.Mempool(
 		ctx,
 		&types.NetworkRequest{
 			NetworkIdentifier: network,
 		},
 	)
 	if err != nil {
-		return nil, fmt.Errorf("%w: /mempool %s", ErrRequestFailed, err.Error())
+		res := &Error{
```
qiwu7, line 39: thoughts about renaming `res` to something like `fetcherError`. This is not really a response I think?

fetcher/utils.go
```diff
@@ -49,3 +50,12 @@ func tryAgain(fetchMsg string, thisBackoff backoff.BackOff, err error) bool {
 
 	return true
 }
+
+// checkError compares a *fetcher.Error to a simple type error and returns
+// a boolean indicating if they are equivalent
+func checkError(fetcherErr *Error, err error) bool {
```
qiwu7, line 56: Can we add some tests for this method?

qiwu7: I guess this is tested implicitly through other existing tests. so should be fine

fetcher/network.go
```diff
@@ -134,52 +158,69 @@ func (f *Fetcher) NetworkListRetry(
...
 		)
-		if errors.Is(err, ErrAssertionFailed) {
-			return nil, fmt.Errorf("%w: /network/list not attempting retry", err)
-		}
-
 		if err == nil {
 			return networkList, nil
 		}
 
+		if errors.Is(err.Err, ErrAssertionFailed) {
```
qiwu7, line 165: nit: maybe we can apply your `checkError` method here

cindyxkuang: hmm, the reason I didn't do it there is because we because we already return the response from our query if `err == nil` (line 162), and `checkError` just safely calls `errors.Is` anyway.

fetcher/mempool.go
cindyxkuang: makes sense, fixed :)

PR #103, opened 2020-08-18
Add support for GetAccountBalances

juliankoh:
Fixes # .

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -279,20 +279,20 @@ func CurrencyBalance(
 	account *types.AccountIdentifier,
 	currency *types.Currency,
 	block *types.BlockIdentifier,
-) (*types.Amount, *types.BlockIdentifier, error) {
+) (*types.Amount, *types.BlockIdentifier, []*types.Coin, error) {
 	var lookupBlock *types.PartialBlockIdentifier
 	if block != nil {
 		lookupBlock = types.ConstructPartialBlockIdentifier(block)
 	}
 
-	liveBlock, liveBalances, _, _, fetchErr := fetcher.AccountBalanceRetry(
+	liveBlock, liveBalances, liveCoins, _, fetchErr := fetcher.AccountBalanceRetry(
 		ctx,
 		network,
 		account,
 		lookupBlock,
 	)
 	if fetchErr != nil {
-		return nil, nil, fmt.Errorf(
+		return nil, nil, nil, fmt.Errorf(
 			"%w: unable to lookup acccount balance for %s",
 			fetchErr.Err,
 			types.PrettyPrintStruct(account),
@@ -301,7 +301,7 @@ func CurrencyBalance(
 
 	liveAmount, err := types.ExtractAmount(liveBalances, currency)
 	if err != nil {
-		return nil, nil, fmt.Errorf(
+		return nil, nil, nil, fmt.Errorf(
 			"%w: could not get %s currency balance for %s",
 			err,
 			types.PrettyPrintStruct(currency),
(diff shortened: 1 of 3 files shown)


storage/balance_storage.go
```diff
@@ -521,3 +521,44 @@ func (b *BalanceStorage) GetAllAccountCurrency(
...
+			accountBalance.Account,
+			accountBalance.BalanceAmount,
+			accountBalance.Block,
+		)
+		if err != nil {
+			return err
+		}
+	}
+
+	err := transaction.Commit(ctx)
```
patrick-ogrady, line 557:
```suggestion
	if err := transaction.Commit(ctx); err != nil {
```

storage/balance_storage.go
```diff
@@ -521,3 +521,44 @@ func (b *BalanceStorage) GetAllAccountCurrency(
...
 	return accounts, nil
 }
+
+// SetBalanceImported sets the balances of a set of addresses by
+// getting their balances from the tip block, and populating the database.
+// This is used when importing prefunded addresses.
+func (b *BalanceStorage) SetBalanceImported(
+	ctx context.Context,
+	helper BalanceStorageHelper,
+	accountBalances []*utils.AccountBalance,
```
patrick-ogrady, line 531: Why create a new type for this? Why not just use `*types.AccountBalanceResponse`?

utils/utils.go
```diff
@@ -301,13 +301,75 @@ func CurrencyBalance(
...
+	Network  *types.NetworkIdentifier
+	Currency *types.Currency
+}
+
+// AccountBalance defines an account's balance,
+// including either balance or coins, as well as
+// the block which this balance was fetched at.
+type AccountBalance struct {
+	Account       *types.AccountIdentifier
+	BalanceAmount *types.Amount
```
patrick-ogrady, line 328: nit: just call this `Amount`

utils/utils.go
```diff
@@ -301,13 +301,75 @@ func CurrencyBalance(
...
+		if len(coins) > 0 {
+			accountBalance = &AccountBalance{
+				Account: balanceRequest.Account,
+				Coins:   coins,
+				Block:   block,
+			}
+		} else {
+			accountBalance = &AccountBalance{
+				Account:       balanceRequest.Account,
+				BalanceAmount: amount,
```
patrick-ogrady, line 366: You need to set this for UTXO chains too

utils/utils.go
```diff
@@ -301,13 +301,75 @@ func CurrencyBalance(
...
+			nil,
+		)
+
+		if err != nil {
+			return nil, err
+		}
+
+		var accountBalance *AccountBalance
+		// Coins exist
+		if len(coins) > 0 {
```
patrick-ogrady, line 357: nit: always set coins, if it doesn't exist it will be nil anyways

storage/coin_storage.go
```diff
@@ -444,3 +445,36 @@ func (c *CoinStorage) GetLargestCoin(
...
+	helper BalanceStorageHelper,
+	accountBalances []*utils.AccountBalance,
+) error {
+	// Update balances in database
+	transaction := c.db.NewDatabaseTransaction(ctx, false)
+	defer transaction.Discard(ctx)
+
+	for _, accountBalance := range accountBalances {
+		for _, coin := range accountBalance.Coins {
+			encodedResult, err := encode(coin)
```
patrick-ogrady, line 463: This loop won't work...adding coins must also update the "wallet" held at each account

patrick-ogrady:
Left a few nits! Looking forward to seeing the tests in `CoinStorage` and `BalanceStorage` to make sure this import works correctly.

storage/coin_storage.go
patrick-ogrady: You can add a test to see this by calling `GetCoins` on each account you add.

utils/utils.go
patrick-ogrady: Balance tracking is always enabled (regardless of if Account-based or UTXO-based)

utils/utils.go
patrick-ogrady: `AccountBalance.BalanceAmount` stutters IMO

utils/utils.go
```diff
@@ -301,13 +301,80 @@ func CurrencyBalance(
...
+	Account *types.AccountIdentifier
+	Amount  *types.Amount
+	Coins   []*types.Coin
+	Block   *types.BlockIdentifier
+}
+
+// GetAccountBalancesHelper is used by GetAccountBalances to determine
+// the CurrencyBalance at tip for an account
+type GetAccountBalancesHelper interface {
+	CurrencyBalance(
```
patrick-ogrady, line 336: nit: it feels weird to define an interface for another function in the same package. I think the better approach would be to replace the `fetcher` call in `CurrencyBalance` with a mock. (as a bonus we would also be testing that function)

utils/utils_test.go
```diff
@@ -188,3 +191,141 @@ func TestRandomNumber(t *testing.T) {
...
+	accBalances, err = GetAccountBalances(
+		ctx,
+		mockHelper,
+		[]*AccountBalanceRequest{accBalanceRequest1},
+	)
+	assert.Nil(t, accBalances)
+	assert.Error(t, err)
+}
+
+type MockFetcherHelper struct {
```
patrick-ogrady, line 290: nit: this isn't how you mock using the stuff you generated...take a look at for an example: https://github.com/coinbase/rosetta-sdk-go/blob/42a93b8550a614ac8d49627afb92ba6cb2774caf/syncer/syncer_test.go#L198-L222

You can remove everything from 290 down and import the mock you generated.

PR #137, opened 2020-09-07
[Fetcher] Improve Connection Reuse

patrick-ogrady:
Blocked by (rebased on): #136

When using the `fetcher` concurrently, it is sometimes possible to surpass the port allocation on the OS. After some research, I determined this was due to not effectively reusing connections. Based on the info in [this thread](https://stackoverflow.com/questions/37774624/go-http-get-concurrency-and-connection-reset-by-peer), I have tuned `IdleConnTimeout`, `MaxIdleConns`, and `MaxIdleConnsPerHost` accordingly.

### Changes
- [x] increase `MaxIdleConns`
- [x] increase `IdleConnTimeout`
- [x] max total connections in fetcher (enforced by semaphore in autogen client) -> default to `120` as most systems have a default max of `128`
- [x] remove `TransactionConcurrency` and set automatically as some % of transactions to fetch (having multiple toggles here is confusing)
- [x] (unrelated) optimize address scan for reconciliation coverage check

### DIFF
--- a/fetcher/account.go
+++ b/fetcher/account.go
@@ -33,6 +33,13 @@ func (f *Fetcher) AccountBalance(
 	account *types.AccountIdentifier,
 	block *types.PartialBlockIdentifier,
 ) (*types.BlockIdentifier, []*types.Amount, []*types.Coin, map[string]interface{}, *Error) {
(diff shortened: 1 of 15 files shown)


juliankoh:
Very nice 🥇

juliankoh (approved):

PR #157, opened 2020-09-15
[Fetcher] Log retry attempts

patrick-ogrady:
When the fetcher retries a request, there is no way to know how many prior attempts have been made. This PR logs this information!

### Changes
- [x] wrap `backoff.BackOff` to count retries

### DIFF
--- a/fetcher/utils.go
+++ b/fetcher/utils.go
@@ -38,15 +38,23 @@ const (
 	clientTimeout = "Client.Timeout exceeded"
 )
 
+// Backoff wraps backoff.BackOff so we can
+// access the retry count (which is private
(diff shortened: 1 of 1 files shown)


fetcher/utils.go
```diff
@@ -89,11 +97,13 @@ func tryAgain(fetchMsg string, thisBackoff backoff.BackOff, err *Error) *Error {
 		errMessage = types.PrintStruct(err.ClientErr)
 	}
 
+	thisBackoff.attempts++
```
patrick-ogrady, line 100: We increment before printing because we initialize at 0.

slowboat0 (approved):

PR #162, opened 2020-09-18
[Storage] Automatically Prune BlockStorage

patrick-ogrady:
Replaces: #161 

This PR includes a method for pruning `block_storage` that does not invalidate any of the [duplicate hash checks](https://github.com/coinbase/rosetta-cli#duplicate-hashes) performed by the `rosetta-cli`.

### Results (> 50% savings 🚀 )
When syncing the first 120k blocks on `bitcoin-testnet`:
```
676M	pruned-bitcoin-data
1.3G	non-pruned-bitcoin-data
```

_Savings will increase dramatically the more block data is ingested (these sizes include all data stored by the `rosetta-cli`, not just block and transaction data)._

### Changes
- [x] implement setter and getter for oldest block index
- [x] track oldest available block in block storage as index
- [x] Implement `Prune` method in `BlockStorage`
- [x] Ensure `SetNewStartIndex` and `CreateBlockCache` error if they try to read a pruned block
- [x] Prevent orphaning past pruned blocks (especially with new start index)
- [x] Throw error if we attempt to access a block that is less than oldest block
- [x] cleanup block storage code
- [x] Attempt to remove old data every X minutes in `Statefulsyncer`

### DIFF
--- a/storage/block_storage.go
+++ b/storage/block_storage.go
@@ -188,22 +188,25 @@ func (b *BlockStorage) pruneBlock(
 	}
 
 	blockResponse, err := b.getBlockResponse(ctx, &types.PartialBlockIdentifier{Index: &oldestIndex}, dbTx)
-	if err != nil {
+	if err != nil && !errors.Is(err, ErrBlockNotFound) {
 		return -1, err
 	}
 
-	blockIdentifier := blockResponse.Block.BlockIdentifier
-	// Set transactions to empty
-	for _, tx := range blockResponse.OtherTransactions {
-		if err := b.pruneTransaction(ctx, dbTx, blockIdentifier, tx); err != nil {
-			return -1, fmt.Errorf("%w: %v", ErrCannotPruneTransaction, err)
+	// If there is an omitted block, we will have a non-nil error.
+	if err == nil {
+		blockIdentifier := blockResponse.Block.BlockIdentifier
+		// Set transactions to empty
+		for _, tx := range blockResponse.OtherTransactions {
+			if err := b.pruneTransaction(ctx, dbTx, blockIdentifier, tx); err != nil {
+				return -1, fmt.Errorf("%w: %v", ErrCannotPruneTransaction, err)
+			}
 		}
-	}
 
-	// Set block to empty
-	_, blockKey := getBlockHashKey(blockIdentifier.Hash)
-	if err := dbTx.Set(ctx, blockKey, []byte(""), true); err != nil {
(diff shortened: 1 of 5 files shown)


storage/block_storage.go
```diff
@@ -104,6 +113,173 @@ func (b *BlockStorage) Initialize(workers []BlockWorker) {
...
+}
+
+// Prune removes block and transaction data
+// from all blocks with index <= index. Pruning
+// leaves all keys associated with pruned data
+// but overwrites their data to be empty. If pruning
+// is successful, we return the range of pruned
+// blocks.
+//
+// Prune is not invoked automatically because
```
cindyxkuang, line 252: Could use some clarification here--what exactly is meant by "invoked automatically"? I see the `Prune()` function in the StatefulSyncer also uses a `for` loop that continuously loops while there is no error, but the comment says that `Prune will repeatedly attempt to prune BlockStorage`

storage/block_storage.go
```diff
@@ -223,18 +403,14 @@ func (b *BlockStorage) GetBlockLazy(
...
 }
 
-// GetBlock returns a block, if it exists. GetBlock
-// will fetch all transactions contained in a block
-// automatically. If you don't wish to do this for
-// performance reasons, use GetBlockLazy.
-func (b *BlockStorage) GetBlock(
+// GetBlockTransactional gets a block in the context of a database
+// transaction.
+func (b *BlockStorage) GetBlockTransactional(
```
cindyxkuang, line 408: The names for the functions `GetBlock` and `GetBlockTransactional` feel like they should be switched to me, or renamed... it's somewhat confusing because `GetBlock` is actually the function invoking a `NewDatabaseTransaction`

storage/block_storage.go
patrick-ogrady: "invoked automatically" here means that `BlockStorage` would start to prune on its own without any invocation. A single call to `Prune()` will prune up to the provided index and then return instead of continuing to prune in the background. `statefulsyncer` calls prune repeatedly to continue to prune the chain.

storage/block_storage.go
patrick-ogrady: Ahh I consider the naming from the perspective of the caller. With `*Transactional`, the caller is responsible for providing the `DatabaseTransaction` to use. Without `Transactional`, the caller is not responsible for providing a transaction where the query will be performed.

slowboat0 (approved):

PR #163, opened 2020-09-18
[Constructor] Load ENVs + Make HTTP Requests

patrick-ogrady:
This PR adds support for reading ENVs and making HTTP requests while running workflows. This would allow anyone to make a request to a faucet during the `request_funds` workflow.

### Changes
- [x] Add support for loading ENVs (`load_env`)
- [x] Make HTTP Requests

### DIFF
--- a/constructor/worker/worker.go
+++ b/constructor/worker/worker.go
@@ -22,6 +22,7 @@ import (
 	"io/ioutil"
 	"log"
 	"net/http"
+	"net/url"
 	"os"
 	"time"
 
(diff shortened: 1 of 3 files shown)


constructor/worker/worker.go
```diff
@@ -673,3 +682,85 @@ func FindCurrencyAmountWorker(rawInput string) (string, error) {
...
+		return "", fmt.Errorf("%w: %s", ErrInvalidInput, err.Error())
+	}
+
+	if input.Timeout <= 0 {
+		return "", fmt.Errorf("%w: %d is not a valid timeout", ErrInvalidInput, input.Timeout)
+	}
+
+	if len(input.URL) == 0 {
+		return "", fmt.Errorf("%w: URL is empty", ErrInvalidInput)
+	}
```
slowboat0, line 715: `url.Parse` can be a better way to validate URL
```suggestion
	if _, err := url.Parse(path); err != nil {
		return "", fmt.Errorf("%w: %s", ErrInvalidInput, err.Error())
	}
```

patrick-ogrady: Ooo very cool. Was not aware!

patrick-ogrady: I opted to go with `ParseRequestURI` instead...it is a lot more restrictive than `Parse`.

slowboat0 (approved):

PR #166, opened 2020-09-22
Assert ConstructionPreprocessResponse.Options not nil

matheusd:
This adds a client assertion to ConstructionPreprocessResponses so that
the Options field is not nil.

The returned options are passed as-is to /construct/metadata and the
server asserter for _that_ endpoint checks that Options isn't nil, so
adding this check makes the error more evident.

### DIFF
--- a/asserter/construction.go
+++ b/asserter/construction.go
@@ -29,6 +29,10 @@ func ConstructionPreprocessResponse(
 		return ErrConstructionPreprocessResponseIsNil
 	}
 
+	if response.Options == nil {
+		return ErrConstructionPreprocessOptionsIsNil
+	}
+
 	for _, accountIdentifier := range response.RequiredPublicKeys {
 		if err := AccountIdentifier(accountIdentifier); err != nil {
 			return err
--- a/asserter/construction_test.go
+++ b/asserter/construction_test.go
@@ -60,6 +60,10 @@ func TestConstructionPreprocessResponse(t *testing.T) {
 		"nil response": {
 			err: ErrConstructionPreprocessResponseIsNil,
 		},
+		"nil options": {
+			response: &types.ConstructionPreprocessResponse{},
+			err:      ErrConstructionPreprocessOptionsIsNil,
+		},
 	}
 
 	for name, test := range tests {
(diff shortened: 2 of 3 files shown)


patrick-ogrady:
👋 @matheusd 

The `ConstructionPreprocessResponse.Options` field is current ["not required" in the specification](https://www.rosetta-api.org/docs/models/ConstructionPreprocessResponse.html). So, adding an assertion here could lead to some incorrect behavior.

If `options` is not populated, the caller should not be invoking `/construction/metadata` and omitting `ConstructionPayloadsRequest.Metadata` in the call to `/construction/payloads`. However, the addition of [suggested fee information to `ConstructionMetadataResponse`](https://www.rosetta-api.org/docs/models/ConstructionMetadataResponse.html) may make this pattern suboptimal (as we wouldn't be able to get suggested fee info). It may be better to make `ConstructionPreprocessResponse.Options` required and add this assertion....

Any thoughts on this point ^^?

matheusd:
Another alternative would be dropping the assertion that Options should be filled in `/construction/metadata` ([here](https://github.com/coinbase/rosetta-sdk-go/blob/master/asserter/server.go#L162)), since it's conceivable you could have a chain that generates no options during `/construction/preprocess` but can generate a suggested fee on `/construction/metadata` (let's say, because that is a network-wide parameter that is independent of any tx details - imagine a blockchain/Rosetta represented fiat payment gateway that charges a fixed fee for txs).

This particular PR was motivated because `rosetta-cli` will pass the `Options` returned by `/construction/preprocess` as-is (during `check:construction`) and isn't currently doing the conditional you mentioned. I'm ok with any of the alternatives that maintain consistency (either `Options` is optional _everywhere_ or it's _always_ required everywhere).

patrick-ogrady:
> This particular PR was motivated because rosetta-cli will pass the Options returned by /construction/preprocess as-is (during check:construction) and isn't currently doing the conditional you mentioned. 
🤦  We'll review this dilemma early next week and respond here.

I agree that we should aim for consistency (either required everywhere or nowhere).

PR #169, opened 2020-09-28
[Utils] Add util for checking if implementation at tip

patrick-ogrady:
To determine if we should begin running `check:construction`, we need to know if we are at tip. This PR adds a convenience function to the `utils` package that does this.

### Changes
- [x] Add a method in `utils` for checking if a Rosetta implementation is at tip using the returned timestamp
- [x] Extend linter timeout


yflinn (dismissed):
🔥

yflinn (approved):

yflinn (approved):

PR #200, opened 2020-10-20
[reconciler] Skip Handler

patrick-ogrady:
This PR ensures no reconciliations "enter the abyss". We now invoke the new `ReconciliationSkipped` handler if we opt to skip reconciliation for any account.

### Changes
- [x] Add `ReconciliationSkipped` hander
- [x] Add `SkippedReconciliationsCounter`
- [x] Cleanup `debugLog`
- [x] Make concurrency variables public

### DIFF
--- a/reconciler/reconciler.go
+++ b/reconciler/reconciler.go
@@ -586,12 +586,24 @@ func (r *Reconciler) accountReconciliation(
 			if errors.Is(err, ErrBlockGone) {
 				// Either the block has not been processed in a re-org yet
 				// or the block was orphaned
+				if r.debugLogging {
+					log.Println(
+						"skipping reconciliation because block gone",
+						types.PrintStruct(liveBlock),
+					)
+				}
 				break
 			}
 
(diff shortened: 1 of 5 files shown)


reconciler/reconciler.go
```diff
@@ -669,37 +753,85 @@ func (r *Reconciler) inactiveAccountQueue(
...
+				}
+
 				continue
 			}
 			if err != nil {
 				// Ensure we don't leak reconciliations if
 				// context is canceled.
 				if errors.Is(err, context.Canceled) {
-					r.wrappedActiveEnqueue(balanceChange)
+					_ = r.wrappedActiveEnqueue(ctx, balanceChange)
```
yflinn, line 834: Looks like the error returned by `wrappedActiveEnqueue` is not actionable. Thoughts on at least logging it instead of skipping?

yflinn (dismissed):

yflinn (approved):

PR #202, opened 2020-10-23
[storage] Historical Balance Tracking

patrick-ogrady:
image

After exposing the number of "skipped" reconciliations in the last release, it became abundantly clear that the `rosetta-cli` wastes up to 50% of all the calls it makes because accounts have been updated before they are able to make an /account/balance request. This PR modifies balance storage to track historical balances for all accounts.

As a side note, this PR unblocks the ability for `rosetta-bitcoin` to serve historical balance queries out of the box!

### Future PR
- [ ] Update `reconciler` package to not do skips
- [ ]  Cleanup BalanceExemption handling in `existingValue` (sometimes rewards are applied before processing, so accounts can go negative)

### DIFF
--- a/storage/balance_storage.go
+++ b/storage/balance_storage.go
@@ -449,12 +449,10 @@ func (b *BalanceStorage) UpdateBalance(
 		)
 		switch {
 		case errors.Is(err, errAccountMissing):
-			fmt.Println("account missing", balance, lastUpdate)
 			storedValue = "0"
 		case err != nil:
 			return err
 		default:
-			fmt.Println("account exists", balance.Value)
 			storedValue = balance.Value
 		}
 
@@ -482,14 +480,10 @@ func (b *BalanceStorage) UpdateBalance(
 		return err
 	}
 
-	fmt.Println("existing value", existingValue)
-
 	newVal, err := types.AddValues(change.Difference, existingValue)
 	if err != nil {
 		return err
 	}
-
-	fmt.Println("new val", newVal)
 
 	bigNewVal, ok := new(big.Int).SetString(newVal, 10)
 	if !ok {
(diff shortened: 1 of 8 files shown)


arjun-io (approved):

storage/balance_storage.go
```diff
@@ -673,3 +816,93 @@ func (b *BalanceStorage) SetBalanceImported(
...
+			if err != nil {
+				return fmt.Errorf(
+					"%w: unable to parse balance entry for %s",
+					err,
+					string(v),
+				)
+			}
+
+			// Ensure block hash matches in case of orphan
+			if deserialBal.Block.Index == block.Index && deserialBal.Block.Hash != block.Hash {
```
qiwu7, line 853: Is it possible that the desired index is not equal to the block index?

arjun-io (approved):

storage/balance_storage.go
patrick-ogrady: Yep! In which case, we will return the balance at the next largest block. This check only makes sure the hashes match if the index does match.

storage/balance_storage.go
```diff
@@ -673,3 +821,93 @@ func (b *BalanceStorage) SetBalanceImported(
...
+				return fmt.Errorf(
+					"wanted block identifier %s but got %s",
+					types.PrintStruct(block),
+					types.PrintStruct(deserialBal.Block),
+				)
+			}
+
+			foundAmount = deserialBal.Amount
+			foundBlock = deserialBal.Block
+			return errAccountFound
```
qiwu7, line 863: hmm why are we returning an error if the account is found? Is it limited by the signature of the Scan method?

patrick-ogrady: This is how we exit the query loop...otherwise we'd need to scan all entries unnecessarily

qiwu7 (approved):

PR #213, opened 2020-10-29
Fetch by Index + Orphan Race Fix

patrick-ogrady:
This PR changes `utils.CurrencyBalance` to fetch by index instead of by hash. This fixes an issue where the `reconciler` requests balances of orphaned blocks (by hash) when processing reconciliations near tip. This PR also addresses a potential race condition where a reorg occurs while comparing balance.

### Changelog
- [x] Change `utils.CurrencyBalance` to only lookup historical balances by index (up to the caller to determine if canonical)
- [x] Get head block, check live block canonical, and get computed balance in same transaction
- [x] Move `reconciler.AccountCurrency` to `types.AccountCurrency` (`storage` should not be depending on `reconciler`)
- [x] Update `ReconcilerHelper` interface to be more transparent about guarantees


qiwu7 (approved):

qiwu7:
@patrick-ogrady looks like there are some failing tests 🤔

patrick-ogrady:
Yep! Sorry this is still WIP in-progress @qiwu7. Was planning a lightweight PR and then decided it was a good opportunity to clean up some other potential races.

reconciler/reconciler.go
```diff
@@ -483,63 +485,36 @@ func (r *Reconciler) bestLiveBalance(
...
-	if errors.Is(err, context.Canceled) || lookupBlock == nil {
-		return nil, nil, liveFetchErr
-	}
-
-	// If there is a reorg, there is a chance that balance
-	// lookup can fail if we try to query an orphaned block.
-	// If this is the case, we continue reconciling.
-	canonical, canonicalErr := r.helper.CanonicalBlock(ctx, lookupBlock)
-	if canonicalErr != nil {
+	if err != nil {
```
patrick-ogrady, line 505: We no longer expect this to error in the case of a re-org.

reconciler/reconciler.go
```diff
@@ -101,27 +102,29 @@ const (
 // what sort of storage layer they want to use to provide the required
 // information.
 type Helper interface {
+	CurrentBlock(
+		ctx context.Context,
+	) (*types.BlockIdentifier, storage.DatabaseTransaction, error)
```
arjun-io, line 107: I don't have full knowledge of how this fits together, but I generally would prefer that whatever is creating the db transaction is also responsible for the commit or rollback. Having one function create a tx and then return it seems a little dangerous since it no longer has any control over the tx. Does the reconciler know enough about `storage` to be able to create the transaction itself and pass it around?

patrick-ogrady: > I don't have full knowledge of how this fits together, but I generally would prefer that whatever is creating the db transaction is also responsible for the commit or rollback.

The storage layer can't handle the commit/rollback because it doesn't know when we are done using this tx (it spans multiple calls). The alternative is to have the reconciler create the transaction and pass it to `CurrentBlock` instead of getting it back.

> Does the reconciler know enough about storage to be able to create the transaction itself and pass it around?

This is a pretty common for packages in `rosetta-sdk-go` for dealing with situations that require database consistency. (I'm not sure there is another way to do these ops without having the packages control the db tx). Take a look at this file for a more complex example of this (it has a separate helper for `DatabaseTransaction` creation): https://github.com/coinbase/rosetta-sdk-go/blob/master/constructor/coordinator/coordinator.go

arjun-io (approved):

arjun-io (approved):

qiwu7 (approved):

PR #216, opened 2020-10-30
[reconciler] Skip on Live Balance Fetch Errors at Tip

patrick-ogrady:
It is possible that the `reconciler` could query an account balance at an index not known to a Rosetta implementation if any balance queries are made before the `syncer` processes a node re-org. This PR ensures we treat these as "skipped" reconciliations instead of failures.

If a failure occurs because of some re-org (and we are querying balance at an index ahead of tip), we will retry the skipped reconciliation when the balance changes from the replacement blocks are added.

### Changes
- [x] Add `IndexAtTip` to `ReconcilerHelper`
- [x] Skip reconciliation if balance fetch fails at tip
- [x] Add `IndexAtTip` to `BlockStorage`

### DIFF
--- a/storage/block_storage.go
+++ b/storage/block_storage.go
@@ -1052,11 +1052,10 @@ func (b *BlockStorage) AtTipTransactional(
 	if errors.Is(err, ErrHeadBlockNotFound) {
 		return false, nil, nil
 	}
-	block := blockResponse.Block
-
 	if err != nil {
 		return false, nil, fmt.Errorf("%w: %v", ErrHeadBlockGetFailed, err)
 	}
+	block := blockResponse.Block
 
 	atTip := utils.AtTip(tipDelay, block.Timestamp)
 	if !atTip {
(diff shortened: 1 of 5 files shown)


storage/block_storage.go
```diff
@@ -1078,3 +1077,54 @@ func (b *BlockStorage) AtTip(
...
+		ctx,
+		&types.PartialBlockIdentifier{Index: &index},
+		transaction,
+	)
+	if err != nil {
+		return false, err
+	}
+	block := blockResponse.Block
+
+	atTip := utils.AtTip(tipDelay, block.Timestamp)
```
arjun-io, line 1124: nit: `return utils.AtTip(tipDelay, blockResponse.Block.Timestamp), nil`

storage/block_storage.go
```diff
@@ -1078,3 +1077,54 @@ func (b *BlockStorage) AtTip(
...
+	if err != nil {
+		return false, err
+	}
+
+	// Check if index is greater than headBlock and if headBlock
+	// is at tip. If so, this is a result of us querying ahead of
+	// tip.
+	headBlock := headBlockResponse.Block
+	if headBlock.BlockIdentifier.Index < index {
+		if utils.AtTip(tipDelay, headBlock.Timestamp) {
```
arjun-io, line 1106: nit: can just do `return utils.AtTip(tipDelay, headBlock.Timstamp), nil`

arjun-io (approved):
Couple of non-blocking nits

storage/block_storage.go
patrick-ogrady: Nice!

patrick-ogrady: Usually my aggressive linter catches stuff like this.

arjun-io (approved):

PR #217, opened 2020-10-31
[reconciler] Fix Pruning Race

patrick-ogrady:
This PR fixes a potential race condition in the optimistic balance pruning logic. This was observed in our own testing when syncing and reconciling with high concurrency.

### Changes
- [x] Reproduce race condition with a test
- [x] Update backlog default to 250k (some networks have congestion spikes)
- [x] Fix race condition
- [x] Add test with multiple changes for same account at same index (different block hashes)
- [x] Increase reconciler safe depth
- [x] Test happy path pruning with 2 calls
- [x] add `BST` implementation to get fast lookup when popular `*types.AccountCurrency` (instead of sorting all keys in a map)

### DIFF
--- a/reconciler/reconciler.go
+++ b/reconciler/reconciler.go
@@ -20,244 +20,12 @@ import (
 	"fmt"
 	"log"
 	"sort"
-	"sync"
 	"time"
 
 	"github.com/coinbase/rosetta-sdk-go/parser"
-	"github.com/coinbase/rosetta-sdk-go/storage"
 	"github.com/coinbase/rosetta-sdk-go/types"
 	"golang.org/x/sync/errgroup"
 )
-
-const (
-	// ActiveReconciliation is included in the reconciliation
-	// error message if reconciliation failed during active
-	// reconciliation.
-	ActiveReconciliation = "ACTIVE"
-
-	// InactiveReconciliation is included in the reconciliation
-	// error message if reconciliation failed during inactive
-	// reconciliation.
-	InactiveReconciliation = "INACTIVE"
-)
-
-const (
-	// BlockGone is when the block where a reconciliation
-	// is supposed to happen is orphaned.
-	BlockGone = "BLOCK_GONE"
-
-	// HeadBehind is when the synced tip (where balances
-	// were last computed) is behind the *types.BlockIdentifier
-	// returned by the call to /account/balance.
-	HeadBehind = "HEAD_BEHIND"
-
-	// BacklogFull is when the reconciliation backlog is full.
-	BacklogFull = "BACKLOG_FULL"
-
-	// TipFailure is returned when looking up the live
-	// balance fails but we are at tip. This usually occurs
-	// when the node processes an re-org that we have yet
-	// to process (so the index we are querying at may be
-	// ahead of the nodes tip).
-	TipFailure = "TIP_FAILURE"
-)
-
-const (
-	// defaultBacklogSize is the limit of account lookups
-	// that can be enqueued to reconcile before new
-	// requests are dropped.
-	defaultBacklogSize = 250000
-
-	// waitToCheckDiff is the syncing difference (live-head)
-	// to retry instead of exiting. In other words, if the
-	// processed head is behind the live head by <
-	// waitToCheckDiff we should try again after sleeping.
-	// TODO: Make configurable
-	waitToCheckDiff = 10
-
-	// waitToCheckDiffSleep is the amount of time to wait
-	// to check a balance difference if the syncer is within
-	// waitToCheckDiff from the block a balance was queried at.
-	waitToCheckDiffSleep = 5 * time.Second
-
-	// zeroString is a string of value 0.
-	zeroString = "0"
-
-	// inactiveReconciliationSleep is used as the time.Duration
-	// to sleep when there are no seen accounts to reconcile.
-	inactiveReconciliationSleep = 1 * time.Second
-
-	// defaultInactiveFrequency is the minimum
-	// number of blocks the reconciler should wait between
-	// inactive reconciliations for each account.
-	defaultInactiveFrequency = 200
-
-	// defaultReconcilerConcurrency is the number of goroutines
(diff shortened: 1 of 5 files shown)


reconciler/reconciler.go
```diff
@@ -331,6 +106,24 @@ func (r *Reconciler) wrappedInactiveEnqueue(
...
+	index int64,
+) {
+	r.queueMapMutex.Lock()
+	key := types.Hash(acctCurrency)
+	if _, ok := r.queueMap[key]; !ok {
+		r.queueMap[key] = map[int64]int{}
+	}
+	if _, ok := r.queueMap[key][index]; !ok {
+		r.queueMap[key][index] = 0
+	}
```
yflinn, line 122: This is not needed `r.queueMap[key][index]` defaults to `0`. I guess the tradeoff here is `better readability` or `slightly better performance` (holding write lock a little bit shorter)

reconciler/reconciler.go
```diff
@@ -331,6 +106,24 @@ func (r *Reconciler) wrappedInactiveEnqueue(
...
 }
 
+// addToqueueMap adds a *types.AccountCurrency
+// to the prune map at the provided index.
+func (r *Reconciler) addToqueueMap(
+	acctCurrency *types.AccountCurrency,
+	index int64,
+) {
+	r.queueMapMutex.Lock()
+	key := types.Hash(acctCurrency)
```
yflinn, line 116: Can we put this outsider of the lock?

```suggestion
	key := types.Hash(acctCurrency)
	r.queueMapMutex.Lock()
```

reconciler/reconciler.go
```diff
@@ -749,6 +548,75 @@ func (r *Reconciler) pruneBalances(ctx context.Context, change *parser.BalanceCh
...
+func (r *Reconciler) updateQueueMapAndPrune(
+	ctx context.Context,
+	change *parser.BalanceChange,
+) error {
+	r.queueMapMutex.Lock()
+
+	key := types.Hash(&types.AccountCurrency{
+		Account:  change.Account,
+		Currency: change.Currency,
+	})
```
yflinn, line 583: Assuming type.Hash() is thread safe
```suggestion
	key := types.Hash(&types.AccountCurrency{
		Account:  change.Account,
		Currency: change.Currency,
	})
	
	r.queueMapMutex.Lock()
```

reconciler/reconciler.go
patrick-ogrady: Was being overly protective to avoid a panic here! Good catch (didn't know that about `int` values): https://play.golang.org/p/-Wb6ZJpsQZU

reconciler/reconciler.go
patrick-ogrady: Yep! It is thread-safe.

reconciler/reconciler.go
```diff
@@ -749,6 +546,75 @@ func (r *Reconciler) pruneBalances(ctx context.Context, change *parser.BalanceCh
...
+		indexes = append(indexes, k)
+	}
+	sort.Slice(indexes, func(i, j int) bool { return indexes[i] < indexes[j] })
+
+	// Don't prune if there are indexes for this AccountCurrency
+	// less than this change.
+	if len(indexes) > 0 && change.Block.Index >= indexes[0] {
+		r.queueMapMutex.Unlock()
+		return nil
+	}
```
yflinn, line 603: Looks like we only care about the `len(indexes)` and `indexes[0]`. We should be able to achieve this without sorting? Not sure about the typical size of the `indexes` but this could be a difference between `O(nlogn)` vs `O(n)` 😛

patrick-ogrady: The size of indexes should be `~= the # of concurrent active reconciliations for a *types.AccountCurrency`. In the worst case, this could be the number of total items in the queue (if the *types.AccountCurrency` is active in each block and reconciliation is super slow).

Let me take another pass at a better algo here!

patrick-ogrady: Alright, I added a BST to make this much faster in the worst case. Shouldn't make much of a difference when there are a handful of indexes for an account but will make a big difference when there are a lot.

utils/bst.go
```diff
@@ -0,0 +1,163 @@
...
+
+// Node is a Node in a BST
+type Node struct {
+	Key   int64
+	Value int
+	left  *Node
+	right *Node
+}
+
+// set adds they key and value to the BST
```
qiwu7, line 65:
```suggestion
// set adds the key and value to the BST
```

reconciler/types.go
```diff
@@ -0,0 +1,258 @@
...
+		ctx context.Context,
+		account *types.AccountIdentifier,
+		currency *types.Currency,
+		index int64,
+	) error
+}
+
+// Handler is called by Reconciler after a reconciliation
+// is performed. When a reconciliation failure is observed,
+// it is up to the client to halt syncing or log the result.
```
qiwu7, line 155:
```suggestion
// it is up to the client to halt syncing or log the result and continue.
```

qiwu7 (approved):

reconciler/reconciler.go
yflinn: Not sure if BST is the right call as it worsens the insertion from O(1) to O(logn) 

What about the following? O(n) to getMin() comparing to O(nlogn) with sort

```
min := func(a, b int) int {
	if a < b {
		return a
	}
	return b
}

minIndex := math.MaxInt64
for k := range r.queueMap[key] {
	minIndex = min(minIndex, k)
}

// Don't prune if there are indexes for this AccountCurrency
// less than this change.
if minIndex != math.MaxInt64 && change.Block.Index >= min {
	r.queueMapMutex.Unlock()
	return nil
}
```

yflinn (approved):

patrick-ogrady:
Resolved @qiwu7's nits! Ready for review @yflinn and @qiwu7 !

yflinn (approved):

PR #221, opened 2020-11-05
[reconciler] Inactive Pruning Race + Inactive Pruning Support

patrick-ogrady:
This PR fixes an inactive pruning race (where the active reconciliation loop prunes a balance being used by the inactive reconciliation loop).

### Changes
- [x] reproduce race condition in tests
- [x] fix race condition
- [x] enable inactive reconciliation pruning (now possible because of race fix)

### DIFF
--- a/reconciler/reconciler.go
+++ b/reconciler/reconciler.go
@@ -111,12 +111,15 @@ func (r *Reconciler) wrappedInactiveEnqueue(
 func (r *Reconciler) addToQueueMap(
 	acctCurrency *types.AccountCurrency,
 	index int64,
-	holdsLock bool,
+	unsafe bool,
 ) {
 	key := types.Hash(acctCurrency)
 
-	if !holdsLock {
+	// We acquire the lock for the queueMap
+	// if the call is safe.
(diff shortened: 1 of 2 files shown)


reconciler/reconciler.go
```diff
@@ -106,15 +106,19 @@ func (r *Reconciler) wrappedInactiveEnqueue(
...
 }
 
-// addToqueueMap adds a *types.AccountCurrency
+// addToQueueMap adds a *types.AccountCurrency
 // to the prune map at the provided index.
-func (r *Reconciler) addToqueueMap(
+func (r *Reconciler) addToQueueMap(
 	acctCurrency *types.AccountCurrency,
 	index int64,
+	holdsLock bool,
```
arjun-io, line 114: I don't hold this opinion too strongly, but instead of having this function conditionally acquire the lock, I would prefer to totally remove the lock acquisition from this function, and instead have the caller acquire/release the lock

arjun-io (dismissed):

arjun-io (approved):

PR #244, opened 2020-11-23
[utils] Implement PriorityMutex

patrick-ogrady:

In an effort to optimize `rosetta-cli` performance, we discovered that a priority lock could be used to prioritize the "hot path" (i.e. the syncing loop). This PR adds a basic implementation of such a lock called `PriorityMutex`.

### Changes
- [x] Add implementation
- [x] Add tests

### DIFF
--- a/utils/priority_mutex.go
+++ b/utils/priority_mutex.go
@@ -1,0 +1,80 @@
+package utils
+
+import (
+	"sync"
+)
+
+// PriorityMutex is a special type of mutex
+// that allows callers to request priority
+// over other callers. This can be useful
+// if there is a "hot path" in an application
+// that requires lock access.
+//
+// WARNING: It is possible to cause lock starvation
+// if not careful (i.e. only high priority callers
+// ever do work).
+type PriorityMutex struct {
+	high []chan struct{}
+	low  []chan struct{}
+
+	m sync.Mutex
+	l bool
+}
+
+// NewPriorityMutex returns a new *PriorityMutex.
+func NewPriorityMutex() *PriorityMutex {
+	return &PriorityMutex{
+		high: []chan struct{}{},
+		low:  []chan struct{}{},
+	}
+}
+
+// Lock attempts to acquire either a high or low
+// priority mutex. When priority is true, a lock
+// will be granted before other low priority callers.
+func (m *PriorityMutex) Lock(priority bool) {
+	m.m.Lock()
+
+	if !m.l {
+		m.l = true
+		m.m.Unlock()
+		return
+	}
+
+	c := make(chan struct{})
+	if priority {
+		m.high = append(m.high, c)
+	} else {
+		m.low = append(m.low, c)
+	}
+
+	m.m.Unlock()
+	<-c
+}
+
+// Unlock selects the next highest priority lock
+// to grant. If there are no locks to grant, it
+// sets the value of m.l to false.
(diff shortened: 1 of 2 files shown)


utils/priority_mutex.go
```diff
@@ -0,0 +1,94 @@
...
+		c := m.low[0]
+		m.low = m.low[1:]
+		close(c)
+		return
+	}
+
+	// We only set m.l to false when there are
+	// no items to unlock because it could cause
+	// lock contention for the next lock to fetch it.
+	m.l = false
```
alex-stone, line 93: Worth adding an assertion that after all of the high and low priority items are processed that we set this back to false at the end of your test?

utils/priority_mutex.go
```diff
@@ -0,0 +1,94 @@
...
+type PriorityMutex struct {
+	high []chan struct{}
+	low  []chan struct{}
+
+	m sync.Mutex
+	l bool
+}
+
+// NewPriorityMutex returns a new *PriorityMutex.
+func NewPriorityMutex() *PriorityMutex {
```
bobg, line 39: Seems to me this constructor is not needed. The zero value of a `PriorityMutex` should serve just fine. (In particular, there's no reason to initialize `high` and `low` to empty slices rather than `nil`.) A caller can just write `pm := new(PriorityMutex)`.

utils/priority_mutex_test.go
```diff
@@ -0,0 +1,69 @@
...
+	time.Sleep(1 * time.Second)
+
+	// Ensure number of expected locks is correct
+	assert.Len(t, l.high, 10)
+	assert.Len(t, l.low, 50)
+
+	l.Unlock()
+	assert.NoError(t, g.Wait())
+
+	// Check array for all high prio, remaining low prio
```
alex-stone, line 67: 💭's on something like:
```suggestion
	// Check results array to ensure all of the high priority items processed first,
	// followed by all of the low priority items. 
```

With all of the abbreviations if you'd like :P, just to clarify that it is checking that the high priority items registered first?

utils/priority_mutex.go
patrick-ogrady: Good call! I've never actually used `append` with a `nil` slice: https://play.golang.org/p/MK0keoRtkJJ

utils/priority_mutex_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"golang.org/x/sync/errgroup"
+)
+
+func TestPriorityMutex(t *testing.T) {
+	arr := []bool{}
+	expected := make([]bool, 60)
+	l := new(PriorityMutex)
```
patrick-ogrady, line 29: Good callout, @bobg !

bobg (approved):

utils/priority_mutex.go
```diff
@@ -0,0 +1,86 @@
...
+//
+// WARNING: It is possible to cause lock starvation
+// if not careful (i.e. only high priority callers
+// ever do work).
+type PriorityMutex struct {
+	high []chan struct{}
+	low  []chan struct{}
+
+	m sync.Mutex
+	l bool
```
itstehkman, line 35: small nit: use more descriptive variable names. `l` is not intuitive to somebody not implementing this

PR #247, opened 2020-11-23
SDK updates for performance, rework BalanceStorage::UpdateBalance

sidhujag:
https://github.com/syscoin/rosetta-syscoin/issues/1, https://github.com/coinbase/rosetta-bitcoin/issues/45 highlights a performance issue related to a block with 80k operations. The design of UpdateBalance causes many async requests to use a db.scan call which is a bottleneck causing +5 hour validation times for the block. This work brings it down to around 15 seconds (even using majority RAM on swap, so likely its better if completely in RAM without thrashing).

I had to introduce the Value in AccountEntry, which will store the latest balance which UpdateBalance requires. This can also replace other places where just the latest balance is needed for an Account as well but I have not done that yet (likely continuing on this PR as we go.)

Tests pass but it is not ready for merge obviously. It is meant for review by @patrick-ogrady and folks over at Coinbase engineering. I hope it either inspires other changes to refactor away from scan calls or further investigations of why scan is slow. I also had thoughts on using multiple iterators using the scan to lookup many changes at once, but the simplicity of this design is a good starting point as I effectively have removed the scan call entirely at the expense of always updating the account table.

Note some other code I updated which is unrelated, I will leave comments directly but I think one might be just an oversight? see the code below under FIles changed.

Disregard the commits, I was essentially debugging quite a bit to find the source of the bottleneck and then worked on alleviating the bottleneck, so you will see a lot of useless commits. I had a deterministic setup to test/fix my changes and determine there is a performance gain.


patrick-ogrady:
Thanks for putting this up, @sidhujag! This is actually how `BalanceStorage` used to work before we added some optimizations to make the `reconciler` work better.

I'm working on a refactor of some other modules to reduce locking contention in general and will make sure to include this change in the next release!

> I hope it either inspires other changes to refactor away from scan calls or further investigations of why scan is slow. I also had thoughts on using multiple iterators using the scan to lookup many changes at once, but the simplicity of this design is a good starting point as I effectively have removed the scan call entirely at the expense of always updating the account table.

I agree with the simplicity here...worth the single extra write IMO (I may break it out into its own key to avoid needing to write `Account` and `Currency` each time the balance is updated). My guess is that the amount of key changes we are making (both deletes and modifications) makes the scan need to iterate over a lot of dead keys/values to get the one we care about.

sidhujag (commented):
> Thanks for putting this up, @sidhujag! This is actually how `BalanceStorage` used to work before we added some optimizations to make the `reconciler` work better.
> 
> I'm working on a refactor of some other modules to reduce locking contention in general and will make sure to include this change in the next release!
> 
> > I hope it either inspires other changes to refactor away from scan calls or further investigations of why scan is slow. I also had thoughts on using multiple iterators using the scan to lookup many changes at once, but the simplicity of this design is a good starting point as I effectively have removed the scan call entirely at the expense of always updating the account table.
> 
> I agree with the simplicity here...worth the single extra write IMO (I may break it out into its own key to avoid needing to write `Account` and `Currency` each time the balance is updated). My guess is that the amount of key changes we are making (both deletes and modifications) makes the scan need to iterate over a lot of dead keys/values to get the one we care about.

Thanks Patrick,

Yea it seems to iterate until it gets to the last value, so if the account has 100 history in that prefix it will iterate 100 times to get to the latest, maybe you can search latest first and break out on the first iteration but I still think its faster this way because its simply not fetching its adding one more field in the account table.

Well putting it into account might make sense anyway. With Syscoin we have a token platform so it makes sense that there would be address, currency and value in an account entry. An address can own multiple assets (currencies). If we integrate our token layer I would guess it would work with that structure.

HelloKashif (commented):
@sidhujag NewDatabaseTransaction doesn't exist anymore on `db.NewDatabaseTransaction(ctx, true)` as it looks like they have changed the interface

sidhujag:
> @sidhujag you'll have to update NewDatabaseTransaction as it doesn't exist anymore on `db.NewDatabaseTransaction(ctx, true)` . Looks like they have changed the interface

that one was locking, great! will review but I think @patrick-ogrady understands the jist of my changes. I think there is a bug in here that was unintentional the namespace change hopefully he also picked that one up.

HelloKashif:
@sidhujag jfyi I tried out your fix in my private fork and tried syncing again. But the improvements weren't that huge.
The sync around 300k blocks is still about 1-2 blocks per second which is almost the same as before.

However I did see that my cpu usage went down quite a lot (from about 12cpu to 4 cpu peak), I guess that's because you stopped using goroutines.

I think we can try adding some more goroutines somewhere in your UpdateBalances function to utilize more cpus. Also as you mentioned not taking a global lock in the db transaction might help

sidhujag:
> @sidhujag jfyi I tried out your fix in my private fork and tried syncing again. But the improvements weren't that huge.
> The sync around 300k blocks is still about 1-2 blocks per second which is almost the same as before.
> 
> However I did see that my cpu usage went down quite a lot (from about 12cpu to 4 cpu peak), I guess that's because you stopped using goroutines.
> 
> I think we can try adding some more goroutines somewhere in your UpdateBalances function to utilize more cpus. Also as you mentioned not taking a global lock in the db transaction might help

makes sense, my fix was around the large changesets per block which isnt the case with bitcoin which is larger changeset across multiple transactions, so I guess the bottleneck is elsewhere. But I think @patrick-ogrady already reworked the code so lets see if it works better for you with his once he is finished his refactor.

patrick-ogrady:
Sorry for the delay on this fix, guys! Ended up needing to do a small refactor to cleanup some testing code. Shooting for EOD Tuesday for a release 🤞

PR #248, opened 2020-11-23
[utils] Implement ShardedMap

patrick-ogrady:

This PR adds support for a `ShardedMap` that enables the caller to concurrently write to a "map-like" data structure (where the caller gets a granular lock based on `key`). 

### Changes
- [x] Implement `ShardedMap`
- [x] Add tests for `ShardedMap`
- [x] Migrate `MutexMap` to use `ShardedMap`

### Future Work
As @bobg mentioned, this `ShardedMap` could be converted to a `ShardedContainer` to allow the caller to store anything they want in a particular shard.


utils/sharded_map.go
```diff
@@ -0,0 +1,85 @@
...
+// shardIndex returns the index of the shard
+// that could contain the key.
+func (m *ShardedMap) shardIndex(key string) int {
+	return int(fnv1a.HashString32(key) % uint32(len(m.shards)))
+}
+
+// Lock acquires the lock for a shard that could contain
+// the key. This syntax allows the caller to perform multiple
+// operations while holding the lock for a single shard.
+func (m *ShardedMap) Lock(key string, priority bool) map[string]interface{} {
```
bobg, line 72: I observe that nothing about this package uses the resulting `map[string]interface{}` _as_ a map. This could conceivably be a more general `ShardedContainer` API, where `Lock` returns an `interface{}` and it's up to the caller to type-assert it to whatever kind of container they're sharding.

This means you'd also have to parameterize the constructor to construct caller-specified container shards rather than assuming they're maps.

One nice benefit of this approach is that (apart from the need to type-assert the result of `Lock`) it's more typesafe, in that the caller can say it's a `map[string]uint64` instead of having to settle for a `map[string]interface{}`.

patrick-ogrady: > I observe that nothing about this package uses the resulting map[string]interface{} as a map. This could conceivably be a more general ShardedContainer API, where Lock returns an interface{} and it's up to the caller to type-assert it to whatever kind of container they're sharding.

With the current API, the only thing that we would probably want to be asserted is that `interface{}` is a pointer. Otherwise, any modifications to the returned object would be ineffectual (or we return a pointer to what is stored, which may get hairy).

> One nice benefit of this approach is that (apart from the need to type-assert the result of Lock) it's more typesafe, in that the caller can say it's a map[string]uint64 instead of having to settle for a map[string]interface{}.

Because the caller (me in `MutexMap`) doesn't need to parse the value in the map if it doesn't exist, this API allows the caller to avoid unnecessary reflection. However, I see your point about why a `ShardedContainer` would be cool.

patrick-ogrady: I think this is a change we could make in a future PR once we have a clear use case in mind (likely just wrapping a `ShardedContainer` implementation). For now, I'd like to keep this optimized for the use case I have in mind (optimization of the `reconciler` and `storage` packages).

storage/badger_storage_test.go
```diff
@@ -309,7 +309,7 @@ func TestBadgerTrain_Limit(t *testing.T) {
 		namespace,
 		newDir,
 		dictionaryPath,
-		10,
+		50,
```
patrick-ogrady, line 312: Increased this to stop a flaky test 😢

utils/mutex_map_test.go
```diff
@@ -68,7 +70,15 @@ func TestMutexMap(t *testing.T) {
 	time.Sleep(1 * time.Second)
 
 	// Ensure number of expected locks is correct
-	assert.Len(t, m.entries, 0)
+	assert.Equal(
```
patrick-ogrady, line 73: 🤮 this is from `make shorten-lines`....going to change this

utils/sharded_map_test.go
```diff
@@ -0,0 +1,69 @@
...
+		assert.Len(t, s, 0)
+		s["test"] = "b"
+		close(a)
+		<-b
+		m.Unlock("b")
+		return nil
+	})
+
+	time.Sleep(1 * time.Second)
+	assert.NoError(t, g.Wait())
```
bobg, line 57: It looks like none of the goroutines in `g` actually need to return an error, which means you could simplify a bit by using [sync.WaitGroup](https://golang.org/pkg/sync/#WaitGroup) instead.

bobg (approved):
One non-blocking suggestion, otherwise LGTM.

utils/sharded_map_test.go
patrick-ogrady: I think adding `sync.WaitGroup` may take more lines/be slightly more complex than just returning `nil`. I think we would need a `wg.Add` before spawning each goroutine and would need to call `wg.Done` at the end of each goroutine. Going to just leave as is for now!

patrick-ogrady: Appreciate the careful review!

PR #251, opened 2020-11-25
[storage] Interface Overhaul

patrick-ogrady:

This PR updates the `storage` interface to provide access to more granular database locking and updates all files in `storage` to use this new interface.

### Future Work
Overhaul `BlockStorage`, `BalanceStorage`, and `CoinStorage` to use `WriteTransaction` to reduce usage of global locking (i.e. pruning/reconciliation don't need to acquire the global database lock).

### DIFF
--- a/storage/badger_storage.go
+++ b/storage/badger_storage.go
@@ -80,7 +80,7 @@ type BadgerStorage struct {
 	encoder  *Encoder
 	compress bool
 
-	writer sync.Mutex
+	writer *utils.MutexMap
 
 	// Track the closed status to ensure we exit garbage
 	// collection when the db closes.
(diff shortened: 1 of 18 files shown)


storage/badger_storage.go
```diff
@@ -171,15 +171,15 @@ func PerformanceBadgerOptions(dir string) badger.Options {
 	// This option will have a significant effect the memory. If the level is kept
 	// in-memory, read are faster but the tables will be kept in memory. By default,
 	// this is set to false.
-	opts.KeepL0InMemory = false
+	opts.KeepL0InMemory = true
```
patrick-ogrady, line 174: In practice, we saw this decrease read/write latency. In the future, we should probably write some benchmarks that allow us to compare db config changes quantitatively.

storage/badger_storage.go
```diff
@@ -197,6 +197,7 @@ func NewBadgerStorage(
 		closed:        make(chan struct{}),
 		pool:          NewBufferPool(),
 		compress:      true,
+		writer:        utils.NewMutexMap(utils.DefaultShards),
```
patrick-ogrady, line 200: I'm going to make this configurable...usually would recommend making this equal to the write concurrency.

yflinn (approved):

PR #252, opened 2020-11-25
[storage] Balance Storage Overhaul

patrick-ogrady:
Closes: #247 

This PR introduces significant optimizations to `BalanceStorage` that dramatically speed up sync speed in the `rosetta-cli` (mostly through the reduction of lock contention). Special shoutout to @yflinn, @sidhujag, and @HelloKashif for debugging the issues that led to this PR.

### Changes
- [x] Use `big.Int` bytes format to store amount
- [x] Don't use `txn.Scan` when updating balance
- [x] cache most recent balance
  - [x] update during orphan
- [x] Don't store pruning index, last reconciled index, and current balance on same record
- [x] Count accounts seen and accounts reconciled
  - [x] Use in `EstimatedReconciliationCoverage`
  - [x] Update accounts reconciled in block syncing loop so that account txs don't conflict
- [x] Rebase on https://github.com/coinbase/rosetta-sdk-go/pull/254
  - [x] squash commits
- [x] Add tests (blocked by https://github.com/coinbase/rosetta-sdk-go/pull/254)
  - [x] Pass existing tests
  - [x] Add test for estimated reconciliation coverage
- [x] Rollback counts when deleting account records (conditionally remove reconciliation if exists)
- [x] Don't error if reconciled orphaned account (could be a race where we delete account while reconciling)

### Future Work (before release)
* limit concurrent goroutines in syncing to `NumCPU` (pretty good default) instead of unbounded (can cause CPU to freeze or OOM)
  * Use: https://github.com/neilotoole/errgroup
* pre-allocate `allChanges` in `parser` (#247 )

PR #255, opened 2020-12-01
[storage] Small Nits

patrick-ogrady:
Related: #252 

This PR completes the `Future Work` section of #252.

### Changes
- [x] use `github.com/neilotoole/errgroup`
- [x] pre-allocate `allChanges`

### DIFF
--- a/parser/balance_changes.go
+++ b/parser/balance_changes.go
@@ -128,9 +128,11 @@ func (p *Parser) BalanceChanges(
 		}
 	}
 
-	allChanges := []*BalanceChange{}
+	i := 0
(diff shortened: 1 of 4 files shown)


parser/balance_changes.go
```diff
@@ -128,9 +128,11 @@ func (p *Parser) BalanceChanges(
 		}
 	}
 
-	allChanges := []*BalanceChange{}
+	i := 0
```
sidhujag, line 131: Ahh I'm not a go expert but I should have seen this that the key is string. ACK this thanks!

yflinn (approved):

PR #258, opened 2020-12-01
[reconciler] Performance Overhaul

patrick-ogrady:

This PR introduces performance optimizations in the `reconciler` that reduce lock contention in the syncing loop. This PR also migrates `BalanceStorage`, `BlockStorage`, and `CoinStorage` to use `github.com/neilotoole/errgroup` (in practice, significantly reduces memory usage without impacting performance).

#### reconciler
- [x] introduce `utils.ShardedMap` to track currently used accounts
- [x] update tests
  - [x] fix failing test where `context.Canceled` returned instead of `reconciliation fails`
- [x] make enqueuing block changes async
#### statefulsyncer
- [x] change `statefulsyncer` default pruning frequency to 10 minutes
#### storage
- [x] explicitly set `errgroup` limit with `runtime.NumCPU`
- [x] ensure `BlockStorage` is not using a global database transaction
- [x] migrate `BlockStorage` to use `github.com/neilotoole/errgroup`
- [x] migrate `CoinStorage` to use `github.com/neilotoole/errgroup`

### DIFF
--- a/reconciler/reconciler.go
+++ b/reconciler/reconciler.go
@@ -29,6 +29,10 @@ import (
 )
 
 const (
+	// shardBuffer is multiplied by inactive concurrency +
+	// active concurrency to determine how many shards should
+	// be created in the queueMap. The more shards created,
+	// the less lock contention we will encounter.
 	shardBuffer = 2
 )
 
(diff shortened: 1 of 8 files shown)


sidhujag:
we are able to sync our node now with official SDK, I will retest again with these changes but so far so good, memory being held in check as well. (We have increased ours due to the database change size issue but can play around there once these changes are in)

yflinn (approved):
LGTM

PR #259, opened 2020-12-01
[chore] Miscellaneous Nits Before Release

patrick-ogrady:
This PR fixes a few miscellaneous issues found while integrating `rosetta-sdk-go` into `rosetta-cli` (https://github.com/coinbase/rosetta-cli/pull/193).

### Changes
- [x] Add `CounterStorage.GetTransactional`
- [x] Fix negative `big.Int` regression: https://play.golang.org/p/wD68BU8AACd
  - [x] find issue in docs: https://golang.org/pkg/math/big/#Int.Bytes
  - [x] reproduce with test
  - [x] don't store pruned if < 0 (otherwise will appear as positive when reading)
- [x] remove erroneous unlock
- [x] delete all historical balances in `deleteAccountRecords`

### DIFF
--- a/storage/modules/balance_storage.go
+++ b/storage/modules/balance_storage.go
@@ -1316,10 +1316,7 @@ func (b *BalanceStorage) removeHistoricalBalances(
 		}
 	}
 
-	// We only update the last pruned index
-	// if we are not orphaning and we actually
-	// deleted some key.
-	if orphan || len(foundKeys) == 0 {
+	if orphan {
 		return nil
 	}
 
--- a/storage/modules/balance_storage.go
+++ b/storage/modules/balance_storage.go
@@ -1317,7 +1317,9 @@ func (b *BalanceStorage) removeHistoricalBalances(
 		}
 	}
 
-	if orphan {
+	// We don't update the pruned value when index is less
+	// than 0 because big.Int conversion doesn't support signed values.
+	if orphan || index < 0 {
 		return nil
 	}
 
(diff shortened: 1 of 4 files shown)


storage/modules/balance_storage.go
```diff
@@ -448,7 +432,7 @@ func (b *BalanceStorage) Reconciled(
 		b.pendingReconciliationMutex.Unlock()
 	}
 
-	if err := dbTx.Set(ctx, key, new(big.Int).SetInt64(block.Index).Bytes(), true); err != nil {
+	if err := dbTx.Set(ctx, key, utils.BigIntBytes(new(big.Int).SetInt64(block.Index)), true); err != nil {
```
patrick-ogrady, line 435: 🤮 this is so ugly and inefficient

sidhujag: why not 
b := make([]byte, 8)
binary.LittleEndian.PutUint64(b, int64(block.Index))  

does bigint do something special to encode the bytes differently?

patrick-ogrady: Probably makes the most sense to just write a `GetInt64` and `StoreInt64` method for some things.

`Amount.Value` would still use `Big.Int`

sidhujag: yea maybe typedef Big.Int as Amount or something and then you can write wrappers and do optimizations as you see fit.

func (a *Amount) AsBigInt() big.Int {
	if a == nil {
		return *new(big.Int)
	}
	return big.Int(*a)
}

func (a *Amount) AsInt64() int64 {
	if a == nil {
		return 0
	}
	return (*big.Int)(a).Int64()
}

patrick-ogrady: Sorry was prototyping some stuff @sidhujag....I actually typed the response to my own message before github showed me yours.

TL;DR I found a way to fix this with a lot less work...

yflinn (approved):

PR #271, opened 2020-12-08
[chore] `rosetta-specifications@v1.4.9`

patrick-ogrady:
This PR updates `rosetta-sdk-go` to the [latest release of `rosetta-specifications`](https://github.com/coinbase/rosetta-specifications/releases/tag/v1.4.9). This is a pretty minor release compared to other recent releases.

### Changelog
- [x] Assert `SyncStatus`
- [x] Assert `SearchTransactionsResponse.TotalCount`

### `rosetta-specifications` Changelog
- Return `Synced` Status [`#71`](https://github.com/coinbase/rosetta-specifications/pull/71)
- [indexer] Add `total_count` field [`#70`](https://github.com/coinbase/rosetta-specifications/pull/70)

### DIFF
--- a/asserter/network.go
+++ b/asserter/network.go
@@ -78,6 +78,27 @@ func Version(version *types.Version) error {
 	return nil
 }
 
+// SyncStatus ensures any types.SyncStatus is valid.
+func SyncStatus(status *types.SyncStatus) error {
+	if status == nil {
+		return nil
+	}
+
+	if status.CurrentIndex != nil && *status.CurrentIndex < 0 {
+		return ErrSyncStatusCurrentIndexNegative
+	}
+
+	if status.TargetIndex != nil && *status.TargetIndex < 0 {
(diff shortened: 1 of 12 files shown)


asserter/network.go
```diff
@@ -78,6 +78,27 @@ func Version(version *types.Version) error {
 	return nil
 }
 
+// SyncStatus ensures any types.SyncStatus is valid.
+func SyncStatus(status *types.SyncStatus) error {
```
septerr, line 82: Since this always returns error (or nil), I'd prefer a name like `SyncStatusError` or `ValidateSyncStatus`. What do you think?

patrick-ogrady: This is consistent with all other functions in `asserter`. Currently, the pattern is that the name of the `asserter` method is the name of the struct it is asserting.

If we want to migrate to your naming strategy, I'd recommend we do that in a separate PR.

septerr (approved):

PR #272, opened 2020-12-08
[storage] Single Errgroup Pool for BlockWorkers

patrick-ogrady:
This PR updates `BlockWorkers` to use a single `errgroup` when adding/removing a block instead of a unique one for each worker (requiring synchronization at the end of each worker spawn).

### Changes
- [x] Update `BlockWorker` interface to accept `errgroup`
- [x] Update `BlockWorker` implementations to use single `errgroup`
- [x] Update tests
- [x] Add argument to block worker initializer to specify the number of BlockWorker executors to run

### DIFF
--- a/mocks/storage/modules/block_worker.go
+++ b/mocks/storage/modules/block_worker.go
@@ -5,6 +5,7 @@ package modules
 import (
 	context "context"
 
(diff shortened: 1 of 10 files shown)


yflinn (approved):
LGTM

PR #273, opened 2020-12-09
[storage] Protect against nil Helper and/or Handler in BalanceStorage

patrick-ogrady:
Someone was able to [force a panic in `rosetta-cli` because of an unpopulated BalanceStorage helper and handler](https://community.rosetta-api.org/t/operation-status-must-be-empty-for-construction/315/4?u=patrick.ogrady). This PR causes BalanceStorage to raise an error in these cases instead of panicking.

### Changes
- [x] Add new error
- [x] Check for nil helper or handler
- [x] Add new tests

### DIFF
--- a/storage/modules/balance_storage.go
+++ b/storage/modules/balance_storage.go
@@ -187,6 +187,10 @@ func (b *BalanceStorage) AddingBlock(
 	block *types.Block,
 	transaction database.Transaction,
 ) (database.CommitWorker, error) {
+	if b.handler == nil {
+		return nil, storageErrs.ErrHelperHandlerMissing
+	}
+
 	changes, err := b.parser.BalanceChanges(ctx, block, false)
 	if err != nil {
 		return nil, fmt.Errorf("%w: unable to calculate balance changes", err)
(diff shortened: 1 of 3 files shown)


storage/modules/balance_storage.go
```diff
@@ -187,6 +187,10 @@ func (b *BalanceStorage) AddingBlock(
 	block *types.Block,
 	transaction database.Transaction,
 ) (database.CommitWorker, error) {
+	if b.handler == nil {
```
septerr, line 190: Can we force the handler to be present in the methods used to create a new BalanceStorage instance?

patrick-ogrady: I wish! There is a somewhat complex dependency process where we can't add the handler or helper during initialization.

patrick-ogrady: We need a reference to `BalanceStorage` to create some useful helpers (we should probably clean this up): https://github.com/coinbase/rosetta-cli/blob/972a1690001f7f7d22fafa0a580512eac8517add/pkg/tester/data.go#L284-L305

septerr: yeah, we should def clean it up!

septerr (approved):

PR #274, opened 2020-12-10
add Hash func (that returns error) to utils

HelloKashif:
### Fixes ###
The `types.Hash` panics on error. Adding a Hash func in utils that returns `(string , error)` .

### Motivation

Current Hash func is a bit hard to use in codebase that don't want to log.Fatal

### Solution

New one returns an error.

### Open questions

We could migrate types.Hash usage in the codebase to utils.Hash in future.

### DIFF
--- a/utils/hash.go
+++ b/utils/hash.go
@@ -1,0 +1,54 @@
+package utils
+
+import (
+	"crypto/sha256"
+	"encoding/json"
+	"fmt"
+)
+
+// hashBytes returns a hex-encoded sha256 hash of the provided
+// byte slice.
+func hashBytes(data []byte) (string, error) {
+	h := sha256.New()
+	_, err := h.Write(data)
+	if err != nil {
+		return "", fmt.Errorf("%w: unable to hash data %s", err, string(data))
+	}
+
+	return fmt.Sprintf("%x", h.Sum(nil)), nil
+}
+
+//NOTE: types package also contains similar Hash() but they panic
+//instead of returning error.
+
+// Hash returns a deterministic hash for any interface.
+// This works because Golang's JSON marshaler sorts all map keys, recursively.
+// Source: https://golang.org/pkg/encoding/json/#Marshal
+// Inspiration:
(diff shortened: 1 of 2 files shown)


patrick-ogrady:
Just curious....have you run into a situation where this panics, @HelloKashif?

I added that `types.Hash` implementation very early on and agree getting that `log.Fatal` out of there is a good idea 😬

utils/hash.go
```diff
@@ -0,0 +1,54 @@
...
+// Hash returns a deterministic hash for any interface.
+// This works because Golang's JSON marshaler sorts all map keys, recursively.
+// Source: https://golang.org/pkg/encoding/json/#Marshal
+// Inspiration:
+// https://github.com/onsi/gomega/blob/c0be49994280db30b6b68390f67126d773bc5558/matchers/match_json_matcher.go#L16
+//
+// It is important to note that any interface that is a slice
+// or contains slices will not be equal if the slice ordering is
+// different.
+func Hash(i interface{}) (string, error) {
```
patrick-ogrady, line 33: I'd like to have all this logic together (potentially having the original `Hash` invoke this function as to not have to maintain 2 identical implementations). Maybe all those `types/utils.go` should be moved into `utils`?

HelloKashif:
> Just curious....have you run into a situation where this panics, @HelloKashif?
> 
> I added that `types.Hash` implementation very early on and agree getting that `log.Fatal` out of there is a good idea 😬

I'm using types.Hash to compare metadata in our custom indexer/search endpoint and wanted to handle errors explicitly.
like 
 ```hash, err:= utils.Hash(somedata)
    if err !=nil {
       //Do our own custom stuff here like closing db, recording stats etc.
    }
````

utils/hash.go
HelloKashif: Yeah definitely, I wasn't sure if I should touch the old one or not. 
I can update this PR to migrate all `types.Hash` to `utils.Hash` . Coming soon!!

PR #278, opened 2020-12-13
feat(keys): support secp256r1

LanfordCai:
### Motivation

This PR is to support `secp256r1` curve for `keys` package

### Solution

Add `ImportPrivateKey`, `GenerateKeypair`, `Sign` and `Verify` functions for `secp256r1` by using golang ecdsa package

### DIFF
--- a/keys/keys.go
+++ b/keys/keys.go
@@ -95,6 +95,9 @@ func ImportPrivateKey(privKeyHex string, curve types.CurveType) (*KeyPair, error
 	case types.Secp256r1:
 		crv := elliptic.P256()
 		x, y := crv.ScalarBaseMult(privKey)
+
+		// IsOnCurve will return false for the point at infinity (0, 0)
+		// See: https://github.com/golang/go/blob/3298300ddf45a0792b4d8ea5e05f0fbceec4c9f9/src/crypto/elliptic/elliptic.go#L24
 		if !crv.IsOnCurve(x, y) {
 			return nil, ErrPubKeyNotOnCurve
 		}
--- a/keys/keys.go
+++ b/keys/keys.go
@@ -15,9 +15,13 @@
 package keys
 
 import (
+	"crypto/ecdsa"
 	"crypto/ed25519"
+	"crypto/elliptic"
+	"crypto/rand"
 	"encoding/hex"
 	"fmt"
+	"math/big"
 
 	"github.com/btcsuite/btcd/btcec"
 
@@ -88,6 +92,24 @@ func ImportPrivateKey(privKeyHex string, curve types.CurveType) (*KeyPair, error
 			PublicKey:  pubKey,
 			PrivateKey: rawPrivKey.Seed(),
 		}
+	case types.Secp256r1:
+		crv := elliptic.P256()
+		x, y := crv.ScalarBaseMult(privKey)
+		rawPubKey := ecdsa.PublicKey{X: x, Y: y, Curve: crv}
+		rawPrivKey := ecdsa.PrivateKey{
+			PublicKey: rawPubKey,
+			D:         new(big.Int).SetBytes(privKey),
+		}
+
+		pubKey := &types.PublicKey{
+			Bytes:     elliptic.Marshal(crv, rawPubKey.X, rawPubKey.Y),
+			CurveType: curve,
+		}
+
+		keyPair = &KeyPair{
+			PublicKey:  pubKey,
+			PrivateKey: rawPrivKey.D.Bytes(),
+		}
 	default:
 		return nil, fmt.Errorf("%w: %s", ErrCurveTypeNotSupported, curve)
 	}
@@ -137,6 +159,22 @@ func GenerateKeypair(curve types.CurveType) (*KeyPair, error) {
 			PublicKey:  pubKey,
 			PrivateKey: rawPrivKey.Seed(),
 		}
+	case types.Secp256r1:
+		crv := elliptic.P256()
+		rawPrivKey, err := ecdsa.GenerateKey(crv, rand.Reader)
+		if err != nil {
+			return nil, fmt.Errorf("%w: %v", ErrKeyGenSecp256r1Failed, err)
+		}
+		rawPubKey := rawPrivKey.PublicKey
+		pubKey := &types.PublicKey{
+			Bytes:     elliptic.Marshal(crv, rawPubKey.X, rawPubKey.Y),
+			CurveType: curve,
+		}
+
+		keyPair = &KeyPair{
+			PublicKey:  pubKey,
+			PrivateKey: rawPrivKey.D.Bytes(),
+		}
 	default:
 		return nil, fmt.Errorf("%w: %s", ErrCurveTypeNotSupported, curve)
 	}
@@ -171,6 +209,8 @@ func (k *KeyPair) Signer() (Signer, error) {
 		return &SignerSecp256k1{k}, nil
 	case types.Edwards25519:
 		return &SignerEdwards25519{k}, nil
+	case types.Secp256r1:
+		return &SignerSecp256r1{k}, nil
 	default:
 		return nil, fmt.Errorf("%w: %s", ErrCurveTypeNotSupported, k.PublicKey.CurveType)
 	}
--- a/keys/keys.go
+++ b/keys/keys.go
@@ -95,6 +95,10 @@ func ImportPrivateKey(privKeyHex string, curve types.CurveType) (*KeyPair, error
 	case types.Secp256r1:
 		crv := elliptic.P256()
 		x, y := crv.ScalarBaseMult(privKey)
+		if !crv.IsOnCurve(x, y) {
+			return nil, ErrPubKeyNotOnCurve
+		}
+
 		rawPubKey := ecdsa.PublicKey{X: x, Y: y, Curve: crv}
 		rawPrivKey := ecdsa.PrivateKey{
 			PublicKey: rawPubKey,
--- a/keys/keys.go
+++ b/keys/keys.go
@@ -97,7 +97,8 @@ func ImportPrivateKey(privKeyHex string, curve types.CurveType) (*KeyPair, error
 		x, y := crv.ScalarBaseMult(privKey)
 
 		// IsOnCurve will return false for the point at infinity (0, 0)
-		// See: https://github.com/golang/go/blob/3298300ddf45a0792b4d8ea5e05f0fbceec4c9f9/src/crypto/elliptic/elliptic.go#L24
+		// See:
+		// https://github.com/golang/go/blob/3298300ddf45a0792b4d8ea5e05f0fbceec4c9f9/src/crypto/elliptic/elliptic.go#L24
 		if !crv.IsOnCurve(x, y) {
 			return nil, ErrPubKeyNotOnCurve
 		}
--- a/keys/signer_secp256r1.go
+++ b/keys/signer_secp256r1.go
@@ -1,0 +1,126 @@
+// Copyright 2020 Coinbase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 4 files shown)


patrick-ogrady:
Thanks for putting this up, @LanfordCai! You'll need to sign your commits before I can merge this 😢

LanfordCai:
Hi, @patrick-ogrady , I have signed my commit :-)

patrick-ogrady:
Waiting for a review internally from some of our cryptography folks, then will merge (if no concerns)!

keys/keys.go
```diff
@@ -88,6 +92,24 @@ func ImportPrivateKey(privKeyHex string, curve types.CurveType) (*KeyPair, error
...
+		}
+
+		pubKey := &types.PublicKey{
+			Bytes:     elliptic.Marshal(crv, rawPubKey.X, rawPubKey.Y),
+			CurveType: curve,
+		}
+
+		keyPair = &KeyPair{
+			PublicKey:  pubKey,
+			PrivateKey: rawPrivKey.D.Bytes(),
```
itstehkman, line 111: Just to double check, is this 32 bytes? That'll be good so that we have uniformity with secp256k1, in which the keypair.PrivateKey is 32 bytes.

itstehkman: Just checked myself, it is 32 bytes. https://play.golang.org/p/n0rt31c3lTs, so we are good here.

keys/signer_secp256r1_test.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	var payloadTests = []payloadTest{
+		{mockPayload(hash("hello123"), types.Ecdsa), types.Ecdsa, 64, false, nil},
+		{
+			mockPayload(hash("hello1234"), types.EcdsaRecovery),
+			types.EcdsaRecovery,
+			65,
+			true,
+			ErrSignUnsupportedSignatureType,
```
itstehkman, line 48: Is ECDSA Recovery truly not supported for secp256r1? Interesting.

keys/keys.go
```diff
@@ -88,6 +92,24 @@ func ImportPrivateKey(privKeyHex string, curve types.CurveType) (*KeyPair, error
 			PublicKey:  pubKey,
 			PrivateKey: rawPrivKey.Seed(),
 		}
+	case types.Secp256r1:
+		crv := elliptic.P256()
+		x, y := crv.ScalarBaseMult(privKey)
+		rawPubKey := ecdsa.PublicKey{X: x, Y: y, Curve: crv}
```
mikelodder7, line 98: Check the `x`,`y` are on the curve and are not zero.

keys/signer_secp256r1.go
```diff
@@ -0,0 +1,126 @@
...
+	privKey := ecdsa.PrivateKey{
+		PublicKey: pubKey,
+		D:         new(big.Int).SetBytes(s.KeyPair.PrivateKey),
+	}
+
+	sigR, sigS, err := ecdsa.Sign(rand.Reader, &privKey, payload.Bytes)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %s", ErrSignFailed, err.Error())
+	}
+	sig := sigR.Bytes()
```
mikelodder7, line 84: check signature is not identity otherwise it will always be valid.

keys/signer_secp256r1.go
```diff
@@ -0,0 +1,126 @@
...
+
+	if err := asserter.Signatures([]*types.Signature{signature}); err != nil {
+		return err
+	}
+
+	message := signature.SigningPayload.Bytes
+	sig := signature.Bytes
+
+	x, y := elliptic.Unmarshal(elliptic.P256(), signature.PublicKey.Bytes)
+	publicKey := ecdsa.PublicKey{X: x, Y: y, Curve: elliptic.P256()}
```
mikelodder7, line 115: Check that the key is not the identity element.

keys/keys.go
itstehkman: How about big endian vs little endian in secp256k1 vs secp256r1?

LanfordCai: Keys and signatures of both curves are encoded in big-endian form

keys/signer_secp256r1_test.go
LanfordCai: This means `type.EcdsaRecovery` sigType for secp256r1 is unsupported for rosetta-go-sdk, not `secp256r1`. ECDSA recovery **IS** supported for `secp256r1`, but his PR would not support this for some reasons:

1. Golang's `crypto/ecdsa` package doesn't provide a `SignRecoverable`  method for `secp256r1` to create recoverable signature({r, s, v}, v is the recovery id). And I didn't find some other long-tested packages provide this. 

2. I know ONT, NEO, ZEEPIN use normal `secp256r1` signatures in their transactions, but I don't know the use case of `secp256r1` recoverable signature. It seems many people just recover the PublicKey from {r, s}, for v is not necessary strictly speaking.

keys/keys.go
LanfordCai: fixed

keys/signer_secp256r1.go
LanfordCai: fixed

keys/signer_secp256r1.go
LanfordCai: This is handled by `ecdsa.Sign` method, please check: https://github.com/golang/go/blob/5a25a3fd1d8248c039e34f5ae02cc9c8198fc6d6/src/crypto/ecdsa/ecdsa.go#L262

keys/keys.go
mikelodder7: Also check that the points x and y are not zero. If they are, IsOnCurve still returns true but we don't want a public key as the identity point since any signature verification will return true.

keys/signer_secp256r1.go
mikelodder7: No that's not it. The signature is composed of two points, R and S. You should check that neither of these is equal to zero aka the identity element on the curve. If the signature is the identity element, it will always be valid.

keys/signer_secp256r1.go
mikelodder7: This checks that the points are on the curve which is excellent but need to also check for the identity element which is when x and y are zero.

keys/keys.go
LanfordCai: > // A Curve represents a short-form Weierstrass curve with a=-3.
//
// Note that the point at infinity (0, 0) is not considered on the curve, and
// although it can be returned by Add, Double, ScalarMult, or ScalarBaseMult, it
// can't be marshaled or unmarshaled, and IsOnCurve will return false for it.

This is the comment of crypto/elliptic, IsOnCurve will return false.

LanfordCai: Please check this: https://github.com/golang/go/blob/3298300ddf45a0792b4d8ea5e05f0fbceec4c9f9/src/crypto/elliptic/elliptic.go#L24 @mikebrowncb

mikebrowncb: 👋 

I think you meant to tag mikelodder

LanfordCai: Oh, I'm sorry

keys/signer_secp256r1.go
LanfordCai: The check for S is here: https://github.com/golang/go/blob/5a25a3fd1d8248c039e34f5ae02cc9c8198fc6d6/src/crypto/ecdsa/ecdsa.go#L272

keys/signer_secp256r1_test.go
itstehkman: Ok, good to know. Better to leave something out if not really used. Less code is more!

patrick-ogrady:
@LanfordCai we are hoping to approve this soon! Sorry for the delay here.

keys/keys.go
itstehkman: Seems valid, and it's tested https://github.com/golang/go/blob/3298300ddf45a0792b4d8ea5e05f0fbceec4c9f9/src/crypto/elliptic/elliptic_test.go#L453

keys/signer_secp256r1.go
```diff
@@ -0,0 +1,140 @@
...
+)
+
+// SignerSecp256r1 is initialized from a keypair
+type SignerSecp256r1 struct {
+	KeyPair *KeyPair
+}
+
+// The Ecdsa signature is the couple (r, s), both r and s are 32 bytes
+const (
+	EcdsaRLen   = 32
```
itstehkman, line 35: small nit: should this be shared with secp256k1? The signing algorithm ECDSA is constant between the two I believe, so these constants should be shared ideally. We can tackle in a future PR though, it's a small nit

keys/signer_secp256r1.go
```diff
@@ -0,0 +1,140 @@
...
+	EcdsaSLen   = 32
+	EcdsaMsgLen = 32
+)
+
+var _ Signer = (*SignerSecp256r1)(nil)
```
itstehkman, line 40: What are you doing with this line? Seems strange and unused.

keys/signer_secp256r1.go
itstehkman: Per above, IsOnCurve checks this

keys/signer_secp256r1.go
patrick-ogrady: This is a common pattern in golang to [verify interface compliance](https://github.com/uber-go/guide/blob/master/style.md#verify-interface-compliance). We use this extensively throughout the SDK and CLI.

keys/signer_secp256r1.go
```diff
@@ -0,0 +1,140 @@
...
+
+// PublicKey returns the PublicKey of the signer
+func (s *SignerSecp256r1) PublicKey() *types.PublicKey {
+	return s.KeyPair.PublicKey
+}
+
+// Sign arbitrary payloads using a KeyPair
```
itstehkman, line 47: nit: document signature format. per below, it seems to be `R || S`

keys/signer_secp256r1.go
itstehkman: Yeah I think you're right, it seems to retry if R or S is 0, but will wait for Mike to ack here.

keys/signer_secp256r1.go
itstehkman: Makes sense. We should add a comment, it's confusing without context

keys/signer_secp256r1.go
LanfordCai: Yes, this is the same in secp256k1, so I named it as EcdsaRLen, not Secp256r1RLen. Maybe I should put this to https://github.com/coinbase/rosetta-sdk-go/blob/23200e54f95116600ff4617be5b78e1768ba689c/keys/signer_secp256k1.go#L33 ? But I don't want to modify secp256k1 code for this PR is for secp256r1, and this is not used in secp256k1 code.

keys/signer_secp256r1.go
LanfordCai: Added a comment for secp256r1 code. For secp256k1 and edwards25519 code, maybe it can be added in a futrue PR.

keys/signer_secp256r1.go
LanfordCai: Updated

keys/signer_secp256r1.go
itstehkman: That's fine, if it's not used in secp256k1 code then we can ignore it or tackle it later. It certainly is a great validation to check sig / message length

keys/signer_secp256r1.go
itstehkman: @mikelodder7 can you re-review?

keys/keys.go
itstehkman: @LanfordCai One more nit, let's just add a comment before the call to IsOnCurve to explicitly show that this is being checked since it is pretty important to know. Then I will +1

LanfordCai: @itstehkman I have added comments before each call to IsOnCurve

itstehkman (dismissed):
Thanks @LanfordCai! This looks great 👏

LanfordCai:
It requires all commits are signed, should I rebase and resign commits and force push to this branch?  @patrick-ogrady

itstehkman:
@LanfordCai yes you should do that. I hope it doesn't wipe out the review history, I don't think it will. But let's see.

LanfordCai:
@itstehkman I have resigned commits and the review history is still there

itstehkman (approved):

patrick-ogrady:
🚀

PR #290, opened 2021-01-08
Add FindRelatedTransactions Function for Block Storage

varunpcb:
### Motivation

This change adds the FindRelatedTransactions function to block storage, which will allow for related transaction processing in the future. This change also adds a duplicate check for related transactions to the asserter.

### Solution
By checking all forward and backward relations, FindRelatedTransactions finds the highest block that must be confirmed to confirm a given transaction.

### DIFF
--- a/asserter/block.go
+++ b/asserter/block.go
@@ -405,7 +405,7 @@ func (a *Asserter) RelatedTransactions(relatedTransactions []*types.RelatedTrans
 	return nil
 }
 
-// ContainsDuplicateCurrency returns nil if no duplicates are found in the array and
+// ContainsDuplicateRelatedTransaction returns nil if no duplicates are found in the array and
 // returns the first duplicated item found otherwise.
 func ContainsDuplicateRelatedTransaction(
 	items []*types.RelatedTransaction,
--- a/asserter/block.go
+++ b/asserter/block.go
@@ -370,6 +370,10 @@ func (a *Asserter) Transaction(
 // any of the related transactions contain invalid types, invalid network identifiers,
 // invalid transaction identifiers, or a direction not defined by the enum.
 func (a *Asserter) RelatedTransactions(relatedTransactions []*types.RelatedTransaction) error {
+	if dup := ContainsDuplicateRelatedTransaction(relatedTransactions); dup != nil {
+		return fmt.Errorf("%w: %v", ErrDuplicateRelatedTransaction, dup)
+	}
+
 	for i, relatedTransaction := range relatedTransactions {
 		if relatedTransaction.NetworkIdentifier != nil {
 			if err := NetworkIdentifier(relatedTransaction.NetworkIdentifier); err != nil {
@@ -396,6 +400,24 @@ func (a *Asserter) RelatedTransactions(relatedTransactions []*types.RelatedTrans
 				i,
 			)
 		}
+	}
+
+	return nil
+}
+
+// ContainsDuplicateCurrency returns nil if no duplicates are found in the array and
+// returns the first duplicated item found otherwise.
+func ContainsDuplicateRelatedTransaction(
+	items []*types.RelatedTransaction,
+) *types.RelatedTransaction {
+	seen := map[string]struct{}{}
+	for _, item := range items {
+		key := types.Hash(item)
+		if _, ok := seen[key]; ok {
+			return item
+		}
+
+		seen[key] = struct{}{}
 	}
 
 	return nil
--- a/asserter/block.go
+++ b/asserter/block.go
@@ -370,7 +370,7 @@ func (a *Asserter) Transaction(
 // any of the related transactions contain invalid types, invalid network identifiers,
 // invalid transaction identifiers, or a direction not defined by the enum.
 func (a *Asserter) RelatedTransactions(relatedTransactions []*types.RelatedTransaction) error {
-	if dup := ContainsDuplicateRelatedTransaction(relatedTransactions); dup != nil {
+	if dup := DuplicateRelatedTransaction(relatedTransactions); dup != nil {
 		return fmt.Errorf("%w: %v", ErrDuplicateRelatedTransaction, dup)
 	}
 
@@ -405,9 +405,9 @@ func (a *Asserter) RelatedTransactions(relatedTransactions []*types.RelatedTrans
 	return nil
 }
 
-// ContainsDuplicateRelatedTransaction returns nil if no duplicates are found in the array and
+// DuplicateRelatedTransaction returns nil if no duplicates are found in the array and
 // returns the first duplicated item found otherwise.
-func ContainsDuplicateRelatedTransaction(
+func DuplicateRelatedTransaction(
 	items []*types.RelatedTransaction,
 ) *types.RelatedTransaction {
 	seen := map[string]struct{}{}
--- a/storage/modules/block_storage.go
+++ b/storage/modules/block_storage.go
@@ -1006,7 +1006,10 @@ func getBackwardRelationKeys(tx *types.Transaction) [][]byte {
 			continue
 		}
 
-		keys = append(keys, getBackwardRelationKey(relatedTx.TransactionIdentifier, tx.TransactionIdentifier))
+		keys = append(
+			keys,
+			getBackwardRelationKey(relatedTx.TransactionIdentifier, tx.TransactionIdentifier),
+		)
 	}
 
 	return keys
@@ -1152,7 +1155,14 @@ func (b *BlockStorage) FindRelatedTransactions(
 		return nil, nil, nil, err
 	}
 
-	for _, child := range children {
+	i := 0
+	for {
+		if i >= len(children) {
+			break
+		}
+		child := children[i]
+		i += 1
+
 		childBlock, childTx, err := b.FindTransaction(ctx, child, db)
 		if err != nil {
 			return nil, nil, nil, err
@@ -1201,7 +1211,7 @@ func (b *BlockStorage) getChildren(
 		func(k []byte, v []byte) error {
 			ss := strings.Split(string(k), "/")
 			txHash := ss[len(ss)-1]
-			txId := &types.TransactionIdentifier{ Hash: txHash }
+			txId := &types.TransactionIdentifier{Hash: txHash}
 			children = append(children, txId)
 			return nil
 		},
--- a/storage/modules/block_storage.go
+++ b/storage/modules/block_storage.go
@@ -1161,7 +1161,7 @@ func (b *BlockStorage) FindRelatedTransactions(
 			break
 		}
 		child := children[i]
-		i += 1
+		i++
 
 		childBlock, childTx, err := b.FindTransaction(ctx, child, db)
 		if err != nil {
@@ -1211,8 +1211,8 @@ func (b *BlockStorage) getChildren(
 		func(k []byte, v []byte) error {
 			ss := strings.Split(string(k), "/")
 			txHash := ss[len(ss)-1]
-			txId := &types.TransactionIdentifier{Hash: txHash}
-			children = append(children, txId)
+			txID := &types.TransactionIdentifier{Hash: txHash}
+			children = append(children, txID)
 			return nil
 		},
 		false,
--- a/storage/modules/block_storage.go
+++ b/storage/modules/block_storage.go
@@ -1216,7 +1216,7 @@ func (b *BlockStorage) FindRelatedTransactions(
 		i++
 
 		// skip duplicates
-		if _, val := seen[childID.Hash]; !val {
+		if _, ok := seen[childID.Hash]; !ok {
 			seen[childID.Hash] = struct{}{}
 		} else {
 			continue
--- a/storage/modules/block_storage.go
+++ b/storage/modules/block_storage.go
@@ -56,6 +56,8 @@ const (
 	blockSyncIdentifier = "blockSyncIdentifier"
 
 	// backwardRelation is a relation from a child to a root transaction
+	// the root is the destination and the child is the transaction listing the root as a backward
+	// relation
 	backwardRelation = "backwardRelation" // prefix/root/child
 )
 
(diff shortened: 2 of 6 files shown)


storage/modules/block_storage.go
```diff
@@ -1084,6 +1126,76 @@ func (b *BlockStorage) FindTransaction(
...
+	var children []*types.TransactionIdentifier
+	for _, relatedTx := range tx.RelatedTransactions {
+		if relatedTx.Direction == types.Forward {
+			children = append(children, relatedTx.TransactionIdentifier)
+		}
+	}
+
+	_, err := db.Scan(
+		ctx,
+		getBackwardRelationKey(tx.TransactionIdentifier, nil),
```
patrick-ogrady, line 1179: This is a lot cleaner than maintaining a separate prefix function. We should probably go back and remove any `...prefix()` functions in another PR.

storage/modules/block_storage.go
```diff
@@ -103,6 +106,17 @@ func getTransactionPrefix(
 	)
 }
 
+func getBackwardRelationKey(
```
patrick-ogrady, line 109: nit: add comment

storage/modules/block_storage.go
```diff
@@ -103,6 +106,17 @@ func getTransactionPrefix(
...
+	root *types.TransactionIdentifier,
+	child *types.TransactionIdentifier,
+) []byte {
+	childHash := ""
+	if child != nil {
+		childHash = child.Hash
+	}
+	return []byte(fmt.Sprintf("%s/%s/%s", transactionNamespace, root.Hash, childHash))
```
patrick-ogrady, line 117: nit: should be `backwardRelation` not `transactionNamespace`

storage/modules/block_storage.go
```diff
@@ -970,6 +992,19 @@ func (b *BlockStorage) storeTransaction(
 	return storeUniqueKey(ctx, transaction, hashKey, encodedResult, true)
 }
 
+func (b *BlockStorage) getBackwardRelationKeys(tx *types.Transaction) [][]byte {
```
patrick-ogrady, line 995: nit: doesn't need to be a method of `BlockStorage`

storage/modules/block_storage.go
```diff
@@ -956,6 +970,14 @@ func (b *BlockStorage) storeTransaction(
 	blockIdentifier *types.BlockIdentifier,
 	tx *types.Transaction,
 ) error {
+	backwardRelationKeys := b.getBackwardRelationKeys(tx)
+	for _, key := range backwardRelationKeys {
+		err := storeUniqueKey(ctx, transaction, key, []byte{}, true)
```
patrick-ogrady, line 975: this won't necessarily be unique unfortunately. We probably just want to do a `transaction.Set`

storage/modules/block_storage.go
```diff
@@ -993,10 +1028,17 @@ func (b *BlockStorage) removeTransaction(
...
 	blockIdentifier *types.BlockIdentifier,
-	transactionIdentifier *types.TransactionIdentifier,
+	tx *types.Transaction,
 ) error {
-	_, hashKey := getTransactionKey(blockIdentifier, transactionIdentifier)
+	backwardRelationKeys := b.getBackwardRelationKeys(tx)
+	for _, key := range backwardRelationKeys {
+		err := transaction.Delete(ctx, key)
+		if err != nil {
+			return fmt.Errorf("error removing backward relation key %s: %w", key, err)
```
patrick-ogrady, line 1037: nit: make this a named error `ErrXXXX`

storage/modules/block_storage.go
```diff
@@ -1084,6 +1126,76 @@ func (b *BlockStorage) FindTransaction(
...
+		if err != nil {
+			return nil, nil, nil, err
+		}
+		children = append(children, newChildren...)
+	}
+
+	return maxBlock, tx, children, nil
+}
+
+func (b *BlockStorage) getChildren(
```
patrick-ogrady, line 1165: nit: add a todo to get children on other networks

storage/modules/block_storage.go
```diff
@@ -1084,6 +1126,76 @@ func (b *BlockStorage) FindTransaction(
...
+	ctx context.Context,
+	tx *types.Transaction,
+	db database.Transaction,
+) ([]*types.TransactionIdentifier, error) {
+	var children []*types.TransactionIdentifier
+	for _, relatedTx := range tx.RelatedTransactions {
```
patrick-ogrady, line 1171: todo: skip if on another network

storage/modules/block_storage.go
```diff
@@ -970,6 +992,19 @@ func (b *BlockStorage) storeTransaction(
 	return storeUniqueKey(ctx, transaction, hashKey, encodedResult, true)
 }
 
+func (b *BlockStorage) getBackwardRelationKeys(tx *types.Transaction) [][]byte {
+	var keys [][]byte
+	for _, relatedTx := range tx.RelatedTransactions {
```
patrick-ogrady, line 997: todo: skip if on another network

storage/modules/block_storage.go
```diff
@@ -1084,6 +1126,76 @@ func (b *BlockStorage) FindTransaction(
 	return newestBlock, newestTransaction, nil
 }
 
+func (b *BlockStorage) FindRelatedTransactions(
+	ctx context.Context,
+	transactionIdentifier *types.TransactionIdentifier,
+	db database.Transaction,
+) (*types.BlockIdentifier, *types.Transaction, []*types.TransactionIdentifier, error) {
+	maxBlock, tx, err := b.FindTransaction(ctx, transactionIdentifier, db)
```
patrick-ogrady, line 1134: nit: check err

patrick-ogrady, line 1134: nit: check max block not nil

storage/modules/block_storage.go
```diff
@@ -1084,6 +1126,76 @@ func (b *BlockStorage) FindTransaction(
...
+		return nil, nil, nil, err
+	}
+
+	for _, child := range children {
+		childBlock, childTx, err := b.FindTransaction(ctx, child, db)
+		if err != nil {
+			return nil, nil, nil, err
+		}
+
+		if maxBlock.Index < childBlock.Index {
```
patrick-ogrady, line 1147: move this check to after check if `childTx` is nil

storage/modules/block_storage.go
patrick-ogrady: if I recall correctly, `nil` NetworkIdentifier implies same network so only skip if not nil and network identifier != network identifier we are syncing

storage/modules/block_storage.go
```diff
@@ -1084,6 +1136,96 @@ func (b *BlockStorage) FindTransaction(
 	return newestBlock, newestTransaction, nil
 }
 
+func (b *BlockStorage) FindRelatedTransactions(
+	ctx context.Context,
+	transactionIdentifier *types.TransactionIdentifier,
+	db database.Transaction,
+) (*types.BlockIdentifier, *types.Transaction, []*types.TransactionIdentifier, error) {
+	maxBlock, tx, err := b.FindTransaction(ctx, transactionIdentifier, db)
```
patrick-ogrady, line 1144: `maxBlock` should be the value of `GetHeadBlockIdentifierTransactional` ... `FindTransaction` just returns the block that the tx was in

storage/modules/block_storage.go
```diff
@@ -1084,6 +1136,96 @@ func (b *BlockStorage) FindTransaction(
...
+		}
+		child := children[i]
+		i++
+
+		childBlock, childTx, err := b.FindTransaction(ctx, child, db)
+		if err != nil {
+			return nil, nil, nil, err
+		}
+
+		if childTx == nil {
```
patrick-ogrady, line 1171: nit: check `childBlock`

storage/modules/block_storage.go
```diff
@@ -1084,6 +1136,96 @@ func (b *BlockStorage) FindTransaction(
...
+	ctx context.Context,
+	transactionIdentifier *types.TransactionIdentifier,
+	db database.Transaction,
+) (*types.BlockIdentifier, *types.Transaction, []*types.TransactionIdentifier, error) {
+	maxBlock, tx, err := b.FindTransaction(ctx, transactionIdentifier, db)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	if maxBlock == nil {
```
patrick-ogrady, line 1149: nit: `rootBlock`

storage/modules/block_storage.go
```diff
@@ -1084,6 +1136,96 @@ func (b *BlockStorage) FindTransaction(
...
+		}
+
+		newChildren, err := b.getChildren(ctx, childTx, db)
+		if err != nil {
+			return nil, nil, nil, err
+		}
+		children = append(children, newChildren...)
+	}
+
+	return maxBlock, tx, children, nil
```
patrick-ogrady, line 1186: Can we de-dup children here (just in case)?

storage/modules/block_storage.go
```diff
@@ -1084,6 +1136,96 @@ func (b *BlockStorage) FindTransaction(
 	return newestBlock, newestTransaction, nil
 }
 
+func (b *BlockStorage) FindRelatedTransactions(
+	ctx context.Context,
+	transactionIdentifier *types.TransactionIdentifier,
+	db database.Transaction,
+) (*types.BlockIdentifier, *types.Transaction, []*types.TransactionIdentifier, error) {
```
patrick-ogrady, line 1143: I'm thinking it makes more sense to return `(*types.BlockIdentifier, *types.Transaction, []*types.Transaction, error)` so the caller doesn't need to re-fetch all related transactions. This won't incur any additional performance overhead for `FindRelatedTransactions` because each call to `FindTransaction` already fetches the `*types.Transaction`.

storage/modules/block_storage.go
varunpcb: which is what we want right? we want to find the latest block we need confirmed to confirm the root transaction. So we find the root block and take the highest index block of all its children.

storage/modules/block_storage.go
```diff
@@ -956,6 +972,14 @@ func (b *BlockStorage) storeTransaction(
 	blockIdentifier *types.BlockIdentifier,
 	tx *types.Transaction,
 ) error {
+	backwardRelationKeys := getBackwardRelationKeys(tx)
+	for _, key := range backwardRelationKeys {
```
patrick-ogrady, line 976: @varunpcb we should only add backwards relation if the transaction exists

storage/modules/block_storage.go
patrick-ogrady: ahhh oops I read that check wrong 😳 ! nvm

storage/modules/block_storage.go
```diff
@@ -1084,6 +1136,108 @@ func (b *BlockStorage) FindTransaction(
...
+		return nil, nil, nil, nil
+	}
+
+	childIds, err := b.getChildren(ctx, tx, db)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	// create map of seen transactions to avoid duplicates
+	seen := make(map[string]bool)
```
patrick-ogrady, line 1159: usually we just do `map[string]struct{}` to create a "set" so that we don't need to explicitly set a value bc it will always be true

storage/modules/block_storage.go
```diff
@@ -970,6 +991,81 @@ func (b *BlockStorage) storeTransaction(
...
+}
+
+func (b *BlockStorage) modifyBackwardRelations(
+	ctx context.Context,
+	transaction database.Transaction,
+	tx *types.Transaction,
+	fn func(ctx context.Context, transaction database.Transaction, key []byte) error,
+) error {
+	var backwardRelationKeys [][]byte
+	for _, relatedTx := range tx.RelatedTransactions {
```
patrick-ogrady, line 1035: do we check for duplicates in the `asserter` for `tx.RelatedTransactions`?

asserter/block.go
```diff
@@ -401,6 +405,24 @@ func (a *Asserter) RelatedTransactions(relatedTransactions []*types.RelatedTrans
 	return nil
 }
 
+// ContainsDuplicateCurrency returns nil if no duplicates are found in the array and
```
cindyxkuang, line 408: nit: change comment to match method name `ContainsDuplicateRelatedTransaction`

storage/modules/block_storage.go
```diff
@@ -1084,6 +1184,108 @@ func (b *BlockStorage) FindTransaction(
...
+	i := 0
+	for {
+		if i >= len(childIds) {
+			break
+		}
+		childID := childIds[i]
+		i++
+
+		// skip duplicates
+		if _, val := seen[childID.Hash]; !val {
```
yflinn, line 1219: nit: thoughts on keeping this checking consistent with the rest of the codebase?
```suggestion
		if _, ok := seen[childID.Hash]; !ok {
```

storage/modules/block_storage.go
```diff
@@ -1084,6 +1184,108 @@ func (b *BlockStorage) FindTransaction(
...
+		func(k []byte, v []byte) error {
+			ss := strings.Split(string(k), "/")
+			txHash := ss[len(ss)-1]
+			txID := &types.TransactionIdentifier{Hash: txHash}
+			children = append(children, txID)
+			return nil
+		},
+		false,
+		false,
+	)
```
septerr, line 1280: Is the purpose of this scan to find transactions where `tx` appears as a backward relation? If so, a comment here would be great! Sorry about dumb asks, this stuff is super confusing to me at first glance 😄

storage/modules/block_storage.go
```diff
@@ -54,6 +54,9 @@ const (
 	// blockSyncIdentifier is the identifier used to acquire
 	// a database lock.
 	blockSyncIdentifier = "blockSyncIdentifier"
+
+	// backwardRelation is a relation from a child to a root transaction
```
septerr, line 58: to add even further clarity, it would be helpful if we mentioned that the root appears in the child's related transactions with a Backward direction. It took me a minute to grasp that at first glance 😅

storage/modules/block_storage.go
```diff
@@ -1084,6 +1184,108 @@ func (b *BlockStorage) FindTransaction(
 	return newestBlock, newestTransaction, nil
 }
 
+func (b *BlockStorage) FindRelatedTransactions(
```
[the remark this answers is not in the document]
varunpcb: this function is meant to fetch all related txs. It currently handles both forward and backward relations. I think the name is fitting as any expansion will require further subroutines regardless.

storage/modules/block_storage.go
varunpcb: no worries definitely a good ask! not a very readable line.

septerr (approved):

PR #291, opened 2021-01-08
Skip reconciliation when interesting account is missing

septerr:
### Motivation

1. There is a bug in our reconciliation logic around missing accounts. When an interesting account is configured, the reconciler may stop with an error like below:

`Error: unable to get computed balance for &{Address:0000000001 SubAccount:<nil>, Metadata:map[]}: &{Symbol: ICP Decimals:8 Metadata:map[]}: block nil`

The reconciler reconciles interesting accounts at every block height. But, there is no entry for an account in balance storage until it is encountered in a block. So, when the reconciler tries to reconcile an interesting account, it hasn't encountered yet, it stops with an account missing error (`ErrAccountMissing`).

2. The error message for `ErrAccountMissing` incorrectly says "block nil". 

### Solution

1. Updated the reconciliation logic to skip reconciliation for missing accounts.
2. Fixed error message.

### DIFF
--- a/reconciler/reconciler.go
+++ b/reconciler/reconciler.go
@@ -24,6 +24,7 @@ import (
 	"golang.org/x/sync/errgroup"
 
 	"github.com/coinbase/rosetta-sdk-go/parser"
+	storageErrors "github.com/coinbase/rosetta-sdk-go/storage/errors"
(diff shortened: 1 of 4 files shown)


patrick-ogrady (approved):
🚀

PR #293, opened 2021-01-12
Utility functions to check storage and network tip status

septerr:
Fixes # .

### Motivation
Rosetta CLI checks to see if it is at tip at a few different places. These checks are not consistent. 

For instance, before constructing transactions, the CLI checks if it is a tip based on (1) if the stored current block is within tip delay or (2) block returned by network/status is at tip (it is within tip delay or has sync status of true) and it is same as the stored current block.

Before broadcasting transactions, the CLI checks if it is at tip based only on if the stored current block is within tip delay. 

These inconsistent checks were leading to the CLI getting stuck during transaction broadcast for chains that use sync status rather than tip delays.

### Solution

Added `CheckStorageTip` and `CheckNetworkTip` functions to SDK utils, to provide clients like the Rosetta CLI a utility method to perform these checks consistently. The CLI would use `CheckStorageTip` to determine if it is at tip.

### Future Work
* Return bool indicating if synced and always return must recent block


utils/utils.go
```diff
@@ -419,26 +500,6 @@ func AtTip(
...
-		ctx,
-		networkIdentifier,
-		nil,
-	)
-	if fetchErr != nil {
-		return false, fmt.Errorf("%w: unable to get network status", fetchErr.Err)
-	}
-
-	return AtTip(tipDelay, status.CurrentBlockTimestamp), nil
-}
```
septerr, line 440: This method is not used by the SDK or the CLI and we are assuming no one else is using it either.

patrick-ogrady:
Places tip is checked in CLI:
* https://github.com/coinbase/rosetta-cli/blob/23e6daf53528a2695b05e6a7f584d4f5e31c1023/pkg/tester/data.go#L501-L521
* https://github.com/coinbase/rosetta-cli/blob/23e6daf53528a2695b05e6a7f584d4f5e31c1023/pkg/tester/construction.go#L329-L347
* https://github.com/coinbase/rosetta-cli/blob/23e6daf53528a2695b05e6a7f584d4f5e31c1023/pkg/processor/broadcast_storage_helper.go#L47-L58

patrick-ogrady (approved):

PR #294, opened 2021-01-12
Update check tip util functions to return both at-tip status and block id

septerr:
Fixes # .

### Motivation

We added CheckNetworkTip and CheckStorageTip utility functions to DRY these checks in the Rosetta CLI. These methods return a block identifier if the network or storage block is at tip respectively. 

In some contexts, the CLI needs to know the block identifier even if it is not at tip.

### Solution

Updated these methods to return at-tip status in addition to the network or storage block identifier.

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -197,8 +197,10 @@ type BlockStorageHelper interface {
 	// to this interface.
 }
 
-// CheckNetworkTip returns block identifier if the block returned by network/status
-// endpoint is at tip. Note that the tipDelay param takes tip delay in seconds.
+// CheckNetworkTip returns a boolean indicating if the block returned by
+// network/status is at tip. It also returns the block identifier
+// returned by network/status.
+// Note that the tipDelay param takes tip delay in seconds.
 // Block returned by network/status is considered to be at tip if one of the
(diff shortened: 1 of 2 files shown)


utils/utils.go
```diff
@@ -208,64 +210,70 @@ func CheckNetworkTip(ctx context.Context,
...
 	}
 
 	// if latest block in storage is not at tip,
 	// check network status
-	tipBlock, fetchErr := CheckNetworkTip(ctx, network, tipDelay, f)
+	networkAtTip, tipBlock, fetchErr := CheckNetworkTip(ctx, network, tipDelay, f)
 	if fetchErr != nil {
-		return nil, fmt.Errorf("%w: unable to fetch network status", fetchErr)
+		return false,
+			currentStorageBlock.BlockIdentifier,
```
patrick-ogrady, line 268: nit: return nil here

septerr: Was on the fence about this. We could leave it to the caller to use or ignore the identifier based on the error. 
But it will be more consistent to return nil in case of errors. Will change.

patrick-ogrady (approved):
🚀

PR #297, opened 2021-01-23
[WIP] Add stack trace to reconciliation errors

yflinn:
Fixes # .


reconciler/reconciler.go
```diff
@@ -282,11 +282,7 @@ func (r *Reconciler) CompareBalance(
 	// Head block should be set before we CompareBalance
 	head, err := r.helper.CurrentBlock(ctx, dbTx)
 	if err != nil {
-		return zeroString, "", 0, fmt.Errorf(
-			"%w: %v",
-			ErrGetCurrentBlockFailed,
-			err,
-		)
+		return zeroString, "", 0, errors.Wrapf(ErrGetCurrentBlockFailed, "%v", err)
```
septerr, line 285: I think this should be `return zeroString, "", 0, errors.Wrapf(err, "%v: %v", ErrGetCurrentBlockFailed.Error() err.Error())`.

This will make sure we are wrapping the `err` and getting its stack trace. And we use `ErrGetCurrentBlockFailed` only for the error message.

PR #302, opened 2021-03-03
Do not override the provided HTTP client in the fetcher

septerr:
### Motivation

When creating a fetcher instance, we always override the `Transport` of the `http.Client` present inside the `APIClient` object. This leads to unexpected behavior when callers provide their own `http.Client` via the `WithClient` option.

### Solution

This change makes it so that we do not override the Transport when it is already provided . Note we still update the Transport's `TLSClientConfig` field when `insecureTLS` is set to true.

### DIFF
--- a/fetcher/fetcher.go
+++ b/fetcher/fetcher.go
@@ -85,13 +85,25 @@ func New(
 	serverAddress string,
 	options ...Option,
 ) *Fetcher {
+	// Override transport idle connection settings
+	//
+	// See this conversation around why `.Clone()` is used here:
+	// https://github.com/golang/go/issues/26013
+	defaultTransport := http.DefaultTransport.(*http.Transport).Clone()
+	defaultTransport.IdleConnTimeout = DefaultIdleConnTimeout
+	defaultTransport.MaxIdleConns = DefaultMaxConnections
+	defaultTransport.MaxIdleConnsPerHost = DefaultMaxConnections
+	defaultHTTPClient := &http.Client{
+		Timeout:   DefaultHTTPTimeout,
+		Transport: defaultTransport,
+	}
+
 	// Create default fetcher
 	clientCfg := client.NewConfiguration(
 		serverAddress,
 		DefaultUserAgent,
-		&http.Client{
-			Timeout: DefaultHTTPTimeout,
-		})
+		defaultHTTPClient,
+	)
 	client := client.NewAPIClient(clientCfg)
 
 	f := &Fetcher{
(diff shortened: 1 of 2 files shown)


fetcher/fetcher.go
```diff
@@ -85,13 +85,25 @@ func New(
...
 	options ...Option,
 ) *Fetcher {
+	// Override transport idle connection settings
+	//
+	// See this conversation around why `.Clone()` is used here:
+	// https://github.com/golang/go/issues/26013
+	defaultTransport := http.DefaultTransport.(*http.Transport).Clone()
+	defaultTransport.IdleConnTimeout = DefaultIdleConnTimeout
+	defaultTransport.MaxIdleConns = DefaultMaxConnections
+	defaultTransport.MaxIdleConnsPerHost = DefaultMaxConnections
```
alex-stone, line 95: The one behavioral change here for clients who are not passing their own rosetta API client is that we're not leveraging the client specified.

We could potentially change this to occur after applying the `options` and check if a client is initialized, e.g.

```
for _, opt := range options {
	opt(f)
}

if f.rosettaClient == nil {
    f.rosettaClient = newDefaultAPIClient(c.maxConnections)
}
```

Also, that way we're also not building out that client, if it is already being passed in with the `WithClient` option.

septerr: Great catch! And great suggestion! There is no need for us to init the fetcher with a client at all. We can delay ituntil after the opts and that gives us much more sensible code!

fetcher/fetcher_test.go
```diff
@@ -195,3 +196,21 @@ func TestInitializeAsserter(t *testing.T) {
...
+	apiClient := client.NewAPIClient(
+		client.NewConfiguration(
+			"https://serveraddress",
+			DefaultUserAgent,
+			httpClient,
+		),
+	)
+	fetcher := New("https://serveraddress", WithClient(apiClient))
+	var assert = assert.New(t)
+	assert.Same(httpClient, fetcher.rosettaClient.GetConfig().HTTPClient)
```
septerr, line 215: lmk if you think we should assert the Transport field of the httpClient remains the same as well.

fetcher/fetcher.go
```diff
@@ -106,18 +96,35 @@ func New(
...
+		}
+
+		// Create default fetcher
+		clientCfg := client.NewConfiguration(
+			serverAddress,
+			DefaultUserAgent,
+			defaultHTTPClient,
+		)
+		client := client.NewAPIClient(clientCfg)
+		f.rosettaClient = client
```
yflinn, line 120: nit: seems like we can consolidate them into (but it's trivial)
```suggestion
		f.rosettaClient = client.NewAPIClient(clientCfg)
```

yflinn (dismissed):
LGTM - left a nit but it's trivial

yflinn (approved):

alex-stone (approved):

heimdall-asguard:
Review Error for alex-stone @ 2021-03-03 02:25:14 UTC
User failed mfa authentication, public email is not set on your github profile. see [go/mfa-help](http://go/mfa-help)

PR #303, opened 2021-03-05
Fix PROTO-1835

raghavsharma873:
Fixes PROTO-1835

### Motivation
A Coinbase dev reported a panic when using Rosetta CLI to view a nonexistent block

### Solution
I think the appropriate change is in rosetta-sdk-go. There is a public method that takes in a pointer to a block. If the caller passes a nil block, the method will panic. Instead, check for a nil block and return an explicit error.

### DIFF
--- a/parser/balance_changes.go
+++ b/parser/balance_changes.go
@@ -36,7 +36,7 @@ type BalanceChange struct {
 // checks indiciating it should be considered a balance change.
 type ExemptOperation func(*types.Operation) bool
 
-var blockNilError = errors.New("block is nil")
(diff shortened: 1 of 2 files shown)


raghavsharma873:
We fixed this in https://github.com/coinbase/rosetta-cli/pull/220
Ideally we would still merge this PR since this extra validation would probably be helpful, but we need to look into how Rosetta error handling works and make sure this doesn't introduce another error class
Either way, I had some Heimdall issues due to misconfiguration so this PR probably isn't mergeable, so deleting it

PR #305, opened 2021-03-10
[Parser] Update parser.MatchOperations to do non-negative and non-positive matching

KNWR:
### Motivation

There are several cases where the matcher may expect opposite amounts but zero amounts are sent. Instead of having clients implement additional fallback logic, we want to update the sdk to be able to handle non-negative or non-positive scenarios when expecting opposite amounts.

### Solution

Add a new OppositeOrZeroAmounts matcher and corresponding matching logic instead of replacing the existing OppositeAmounts matcher. This way, this isn't a breaking change for other users of the library.

Add PositiveAmountSignOrZero and NegativeAmountSignOrZero amount signs for amount matching when using  `OppositeOrZeroAmounts`. We could just use AnyAmountSign, but these additional signs let us get the same order of matches as with `OppositeAmounts`. 

### Open questions

1. Does the order of elements in the Matches array returned by `MatchOperations` matter?


parser/match_operations_test.go
```diff
@@ -105,6 +105,79 @@ func TestMatchOperations(t *testing.T) {
...
+			descriptions: &Descriptions{
+				OppositeOrZeroAmounts: [][]int{{0, 1}},
+				OperationDescriptions: []*OperationDescription{
+					{
+						Account: &AccountDescription{
+							Exists: true,
+						},
+						Amount: &AmountDescription{
+							Exists: true,
+							Sign:   AnyAmountSign,
```
KNWR, line 137: Test passes when this is NegativeAmountSign and L146 is PositiveAmountSign - not sure why

KNWR: Actually - if I reverse the order of the Matches output expected for this test (i.e. the below), still using AnyAmountSign, it passes. This is also true of the other test I added (with 0 amount values). I'm unsure if changing the order changes the correctness of the result?

```
			matches: []*Match{
				{
					Operations: []*types.Operation{
						{
							Account: &types.AccountIdentifier{
								Address: "addr2",
							},
							Amount: &types.Amount{
								Value: "100",
							},
						},
					},
					Amounts: []*big.Int{big.NewInt(100)},
				},
				{
					Operations: []*types.Operation{
						{
							Account: &types.AccountIdentifier{
								Address: "addr1",
							},
							Amount: &types.Amount{
								Value: "-100",
							},
						},
					},
					Amounts: []*big.Int{big.NewInt(-100)},
				},
			},
			err: false,
		},
```

parser/match_operations_test.go
```diff
@@ -105,6 +105,152 @@ func TestMatchOperations(t *testing.T) {
 			},
 			err: false,
 		},
+		"simple transfer (with OppositeOrZeroAmounts) and opposite value amounts": {
```
KNWR, line 108: For both added tests, the code returns the expected Matches output, but in reverse order (i.e. the below passes for the first added test, similarly reversed output passes for the second). Are the matches with this changed order still a correct output?

Possibly relevant: the code passes the first test and produces non-reversed Matches when using NegativeAmountSign and PositiveAmountSign instead of AnyAmountSign, so the sign may impact this ordering.

```
			matches: []*Match{
				{
					Operations: []*types.Operation{
						{
							Account: &types.AccountIdentifier{
								Address: "addr2",
							},
							Amount: &types.Amount{
								Value: "100",
							},
						},
					},
					Amounts: []*big.Int{big.NewInt(100)},
				},
				{
					Operations: []*types.Operation{
						{
							Account: &types.AccountIdentifier{
								Address: "addr1",
							},
							Amount: &types.Amount{
								Value: "-100",
							},
						},
					},
					Amounts: []*big.Int{big.NewInt(-100)},
				},
			},
			err: false,
		},
```

parser/match_operations.go
```diff
@@ -28,13 +28,19 @@ type AmountSign int
...
 	// NegativeAmountSign is a negative amount.
-	NegativeAmountSign
+	NegativeAmountSign = 1
 
 	// PositiveAmountSign is a positive amount.
-	PositiveAmountSign
+	PositiveAmountSign = 2
+
+	// PositiveOrZeroAmountSign is a positive or zero amount.
+	PositiveAmountSignOrZero = 3
```
septerr, line 40: Comment (`PositiveOrZeroAmountSign `) and constant names (`PositiveAmountSignOrZero `) don't match.

I would personally prefer `PositiveOrZeroAmountSign`, for naming consistency, even though zero is not a sign.

parser/match_operations.go
```diff
@@ -28,13 +28,19 @@ type AmountSign int
...
+
+	// NegativeOrZeroAmountSign is a positive or zero amount.
+	NegativeAmountSignOrZero = 4
```
septerr, line 43: Same comment as before :)

parser/match_operations.go
```diff
@@ -129,17 +143,22 @@ type OperationDescription struct {
...
+	// OppositeAmounts are specified using the operation indices of
 	// OperationDescriptions to handle out of order matches. MatchOperations
 	// will error if all groups of operations aren't opposites.
 	OppositeAmounts [][]int
 
-	// EqualAddresses are specified using the operation indicies of
+	// OppositeZeroAmounts are specified using the operation indices of
+	// OperationDescriptions to handle out of order matches. MatchOperations
+	// will error if all groups of operations aren't 0 or opposites.
+	OppositeOrZeroAmounts [][]int
```
septerr, line 159: Name in comment (`OppositeZeroAmounts `) and constant name (`OppositeOrZeroAmounts `) do not match.

parser/match_operations.go
```diff
@@ -482,44 +535,55 @@ func comparisonMatch(
...
-			return fmt.Errorf("cannot check opposites of %d operations", len(amountMatch))
-		}
+	oppositeAmountMatchingOperations := [][][]int{
+		descriptions.OppositeOrZeroAmounts,
+		descriptions.OppositeAmounts,
+	}
+	oppositeAmountCheckers := []func(*types.Operation, *types.Operation) error{
+		oppositeOrZeroAmounts,
+		oppositeAmounts,
+	}
```
septerr, line 545: The 3d array makes grasping the code kinda difficult (this file is pretty difficult to read as is :( ). 

Can we instead utilize a function to which we pass `amountPairs`, `amountChecker`  and anything else that is needed (`matches`) and call it twice. Once for `OppositeAmounts ` and once for `OppositeOrZeroAmounts `? 🙏

septerr (commented):
looks great! left some smoll comments! 🙏

septerr (dismissed):

septerr (approved):

septerr (approved):

PR #313, opened 2021-03-29
add logger for insufficiently funded accounts during construction

lenny-hardenberg:
### Motivation
When running Rosetta CLI `check:construction` we might run into an infinite loop without any clear logs that tell us why we are in a loop. Specifically, this might happen when we are looking for a wallet/account with a certain balance, but none of our `prefunded_accounts` have sufficient funds.

### Solution
Log "[number] account(s) insufficiently funded. Did you forget to fund? [list of addresses]" whenever we call `find_balance` and we are unable to match any balance.

### DIFF
--- a/constructor/worker/worker.go
+++ b/constructor/worker/worker.go
@@ -646,6 +646,7 @@ func (w *Worker) FindBalanceWorker(
 		return "", ErrCreateAccount
 	}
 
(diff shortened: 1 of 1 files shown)


yflinn (approved):
LGTM. cc @septerr

septerr (approved):

PR #315, opened 2021-04-05
249

### LINKED ISSUE #249: TestPriorityMutex fails under `go test -race`
**Describe the bug**
The Go race detector finds a race condition when running `TestPriorityMutex`.

**To Reproduce**
`go test -race ./utils`

**Expected behavior**
No race condition found.

raghavsharma873:
Fixes #249  .

### Solution
Don't rely on `time.Sleep` since there are no formal guarantees
Instead use a WaitGroup-based wrapper so we can properly assert that all goroutines
have asked for lock

### DIFF
--- a/utils/priority_mutex_test.go
+++ b/utils/priority_mutex_test.go
@@ -85,7 +85,7 @@ func TestPriorityMutex(t *testing.T) {
 	assert.Len(t, l.low, 50)
 
 	l.Unlock()
-	assert.NoError(t, g.Wait()) // Wait for all goroutines to ask for lock
+	assert.NoError(t, g.Wait())
 
 	// Check results array to ensure all of the high priority items processed first,
 	// followed by all of the low priority items.
--- a/utils/priority_mutex_test.go
+++ b/utils/priority_mutex_test.go
@@ -16,10 +16,11 @@ package utils
 
 import (
 	"context"
-	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


utils/priority_mutex_test.go
```diff
@@ -16,17 +16,41 @@ package utils
...
+func newWrappedPriorityMutex(numLockCalls int) *wrappedPriorityMutex {
+	w := &wrappedPriorityMutex{PriorityMutex: new(PriorityMutex)}
+	w.lockWg.Add(numLockCalls)
+	return w
+}
+
+func (w *wrappedPriorityMutex) Lock(priority bool) {
+	c := w.PriorityMutex.lockInternal(priority)
+	w.lockWg.Done()
+	w.wait(c)
```
septerr, line 43: In this method, if we do following, I think we can initialize the WaitGroup counter with 60 instead of 61. 60 will make the test easier to understand. lmk what you think. 🙏 
```
func (w *wrappedPriorityMutex) Lock(priority bool) {
	c := w.PriorityMutex.lockInternal(priority)
        if c != nil {
          w.lockWg.Done()
	  w.wait(c)
        }
}
```

septerr (approved):
🎉

raghavsharma873:
I don't have write access to this repo so I can't merge. Feel free to merge on my behalf. I'm requesting access

PR #316, opened 2021-04-06
Fix 239

### LINKED ISSUE #239: MathOperation : Support Multiply / Division Operation
Calculation on balance need more operations. Now, only **addition** and **subtraction** are supported. It's great to have **multiply** and **division** 

Support Support Multiply / Division Operation https://github.com/coinbase/rosetta-sdk-go/blob/1e2b4a3d02e14ed747e02fefc9cfabb9e36c3418/constructor/job/types.go#L171

raghavsharma873:
Fixes #239  .

### Motivation
Support multiplication and division in the Rosetta DSL

### Solution
It's quite similar to supporting addition and subtraction
Basically just add boilerplate that wraps around bigint multiplication and division
I added tests for the DSL syntax and the multiplication and division logic


types/utils.go
```diff
@@ -143,6 +143,46 @@ func SubtractValues(
...
+
+	bVal, err := BigInt(b)
+	if err != nil {
+		return "", err
+	}
+
+	newVal := new(big.Int).Div(aVal, bVal)
+	return newVal.String(), nil
+}
+
```
septerr, line 185: Can we add tests for these in `utils_test.go`. 🙏

septerr (approved):

PR #324, opened 2021-06-21
fix error wrap

ZhangxyCC:
### Motivation

The current `RequestFailedError` function will return a new error which is not the same type as the initial error. 
### Solution

pass the error into `fmt.Errorf` instead of `err.Error()`

### DIFF
--- a/fetcher/errors.go
+++ b/fetcher/errors.go
@@ -55,7 +55,7 @@ func (f *Fetcher) RequestFailedError(
 	}
 
 	return &Error{
-		Err:       fmt.Errorf("%w: %s %s", ErrRequestFailed, message, err.Error()),
+		Err:       fmt.Errorf("%w: %s %s", ErrRequestFailed, message, err),
 		ClientErr: rosettaErr,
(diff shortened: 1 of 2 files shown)


fetcher/errors.go
```diff
@@ -55,7 +55,7 @@ func (f *Fetcher) RequestFailedError(
 	}
 
 	return &Error{
-		Err:       fmt.Errorf("%w: %s %s", ErrRequestFailed, message, err.Error()),
+		Err:       fmt.Errorf("%w: %s %s", ErrRequestFailed, message, err),
```
septerr, line 58: did you mean to change this to: `fmt.Errorf("%w: %s %s", err, message, ErrRequestFailed)`? Mapping %w to original error, will preserve the error chain.

septerr (commented):
it would be great if can we add a test for this change 🙏

fetcher/errors.go
ZhangxyCC: exactly, will add test asap!

PR #326, opened 2021-08-18
Introduce a validation config for validating operation using rosetta-cli

shrimalmadhur:
Fixes # .

### Motivation
There are lots of validation `rosetta-cli` doesn't perform on operation. 
Few of them include
- validating that deposit and withdraw balances
- issuers might want to validate the number of operation for payment or fee

### Solution
We introduce a validation config file which an asset issuer can specify and provide as an argument while doing `check:data` (more to come on how to use)

TODO:
- Add more tests
- Figure out how to release the package so that it doesn't break existing usage


asserter/asserter.go
```diff
@@ -147,6 +186,9 @@ func NewClientWithFile(
 		config.AllowedOperationStatuses,
 		config.AllowedErrors,
 		&config.AllowedTimestampStartIndex,
+		&Validations{
+			Enabled: false,
```
xiaying-peng, line 190: Could you add a comment why we do not need validation config here?

asserter/block.go
```diff
@@ -293,12 +293,31 @@ func (a *Asserter) Operations(
...
+	zero := big.NewInt(0)
 	for i, op := range operations {
 		// Ensure operations are sorted
 		if err := a.Operation(op, int64(i), construction); err != nil {
 			return err
 		}
 
+		if a.validations.Enabled {
+			if op.Type == a.validations.Payment.Name {
+				val, _ := new(big.Int).SetString(op.Amount.Value, 10)
```
xiaying-peng, line 309: I think the result boolean matters, which indicate conversion succeed or not.

asserter/block.go
```diff
@@ -293,12 +293,31 @@ func (a *Asserter) Operations(
...
+				paymentTotal.Add(paymentTotal, val)
+				paymentCount++
+			}
+
+			if op.Type == a.validations.Fee.Name {
+				val, _ := new(big.Int).SetString(op.Amount.Value, 10)
```
xiaying-peng, line 315: same here, fail if unable to convert.

asserter/block_test.go
```diff
@@ -268,6 +268,134 @@ func TestAccountIdentifier(t *testing.T) {
...
+		}
+	)
+	var tests = map[string]struct {
+		operations         []*types.Operation
+		validationFilePath string
+		asserter           *Asserter
+		construction       bool
+		err                error
+	}{
+		"valid operations based on validation file": {
```
xiaying-peng, line 307: Add failure cases. Also bad amount cases: make to_string conversion failed.

fetcher/fetcher_test.go
```diff
@@ -189,7 +189,7 @@ func TestInitializeAsserter(t *testing.T) {
 				WithRetryElapsedTime(5*time.Second),
 			)
 
-			networkIdentifier, networkStatus, err := f.InitializeAsserter(ctx, test.network)
+			networkIdentifier, networkStatus, err := f.InitializeAsserter(ctx, test.network, "")
```
xiaying-peng, line 192: cannot believe Golang does not support default value..

asserter/asserter.go
```diff
@@ -261,3 +305,20 @@ func (a *Asserter) OperationSuccessful(operation *types.Operation) (bool, error)
 
 	return val, nil
 }
+
+func getValidationConfig(validationFilePath string) (*Validations, error) {
+	validationConfig := &Validations{
+		Enabled: false,
```
xiaying-peng, line 311: should we enable it by default?

asserter/block.go
shrimalmadhur: We are already verifying this with the `Operation` call at line 303. 
https://github.com/coinbase/rosetta-sdk-go/blob/master/asserter/block.go#L53 - so we don't need it here

asserter/block.go
shrimalmadhur: We are already verifying this with the `Operation` call at line 303. 
https://github.com/coinbase/rosetta-sdk-go/blob/master/asserter/block.go#L53 - so we don't need it here

asserter/asserter.go
shrimalmadhur: I think to support backward compatibility, it makes sense to disable it by default.

fetcher/fetcher_test.go
shrimalmadhur: 🤷

asserter/asserter.go
shrimalmadhur: `rosetta-cli` doesn't use this function. I tried to limit this functionality to that so that's why disabled it.

xiaying-peng (approved):
LGTM!

PR #327, opened 2021-08-19
upgrade to go 1.16 to fix addlicense library and build

shrimalmadhur:
Fixes # .

### Motivation
Due to https://github.com/google/addlicense/commit/d43bb61fdfdafb29f4b1add4b849c5bfe4eeb497 The addlicense functionality was failing. So the only way was to upgrade everything to 1.16 

### Solution
- upgraded all go images to 1.16
- changed the installation commands of `addlicense`, `golines` and `coveralls`
- using `go install` instead of `go run` since `go run` doesn't download the libraries anymore - https://go.dev/blog/go116-module-changes


Makefile
```diff
@@ -51,16 +54,20 @@ check-format:
 test:
 	${TEST_SCRIPT}
 
-test-cover:	
+test-cover:
+	${GOVERALLS_INSTALL}
```
septerr, line 58: out of curiosity, why did you decide to change from `go run` to first installing the binaries?

shrimalmadhur: Good question. So I have been getting this error on the CI [builds](https://app.circleci.com/pipelines/github/coinbase/rosetta-sdk-go/2018/workflows/d8f5c728-e102-4dcd-ac97-2038b3cf6b83/jobs/14862) and on the local when I directly run `go run`
```
go run github.com/google/addlicense -c "Coinbase, Inc." -l "apache" -v .
no required module provides package github.com/google/addlicense; to add it:
	go get github.com/google/addlicense
```

Now I tried doing `go get ` and updating the `go.mod` but since the package doesn't use those libraries, `[go mod tidy](https://github.com/coinbase/rosetta-sdk-go/blob/master/codegen.sh#L212)` ends up cleaning those packages from `go.mod`

shrimalmadhur: Also I think this is because of new updates to `go 1.16` - https://go.dev/blog/go116-module-changes
Go command won't explicitly download the dependency if missed. So for these specific libraries which are not used in main package, we can do `go install` and run them.

septerr (approved):

PR #329, opened 2021-08-27
Improve logging when unable to lookup balance > min

xiaying-peng:
### Motivation
During `rosetta-cli check:construction`, when accounts does not have enough fund, or minimal required balance is set too high in `ros` config, it is hard to debug and diagnose. 

### Solution
Add extra debug logging when we are unable to lookup for amounts > min.


investorguddi (commented):
True

PR #354, opened 2021-11-16
Check related ops

ash-krishnan:
Fixes # .

### Motivation

Currently we don't flag an asset if it doesn't have related operations implemented. Implementing related operations is optional, we only need it to be implemented if we have multiple outputs or we don't distinguish between sends/receives. In that case, ChainIO will break because it uses related operations to enforce a DAG structure on the operations.

### Solution

Make related operations support configurable. If a related_operations_enabled is set to true in asserter file, then throw error if related operations key doesn't exist. Otherwise, just print warning.

### DIFF
--- a/asserter/asserter.go
+++ b/asserter/asserter.go
@@ -40,6 +40,7 @@ type Asserter struct {
 	callMethods             map[string]struct{}
 	mempoolCoins            bool
 	validations             *Validations
+	relatedOpsEnabled       bool
 }
 
 // Validations is used to define stricter validations
@@ -78,6 +79,7 @@ func NewServer(
 	callMethods []string,
 	mempoolCoins bool,
 	validationFilePath string,
+	relatedOpsEnabled bool,
 ) (*Asserter, error) {
 	if err := OperationTypes(supportedOperationTypes); err != nil {
 		return nil, err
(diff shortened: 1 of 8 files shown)


asserter/asserter.go
```diff
@@ -78,6 +79,7 @@ func NewServer(
 	callMethods []string,
 	mempoolCoins bool,
 	validationFilePath string,
+	relatedOpsEnabled bool,
```
shrimalmadhur, line 82: We should put this as a part of validation config rather than create a new parameter. This breaks the existing usage.

ash-krishnan: so we want to add this to the validations struct along with `enabled`, `chainType` etc.?

shrimalmadhur: Yep. the reason we added that validation file is to use and extend with more functionalities.

ash-krishnan: got it! we'd only be able to check/throw an error for account based chains then

ash-krishnan: opened a new PR: https://github.com/coinbase/rosetta-sdk-go/pull/355

shrimalmadhur: I mean we could so this check on both. Right now we only do for account based but we can change the code to have for both UTXO and account.

PR #355, opened 2021-11-16
Related ops checks

ash-krishnan:
Fixes # .

### Motivation

Currently we don't flag an asset if it doesn't have related operations implemented. Implementing related operations is optional, we only need it to be implemented if we have multiple outputs or we don't distinguish between sends/receives. In that case, ChainIO will break because it uses related operations to enforce a DAG structure on the operations.

### Solution

Make related operations support configurable. If a related_operations_enabled is set to true in validator file, then throw error if related operations key doesn't exist. Otherwise, just print warning.

### DIFF
--- a/asserter/asserter.go
+++ b/asserter/asserter.go
@@ -46,11 +46,11 @@ type Asserter struct {
 // on the transaction. Fore more details please refer to
 // https://github.com/coinbase/rosetta-sdk-go/tree/master/asserter#readme
 type Validations struct {
-	Enabled    bool                 `json:"enabled"`
-	RelatedOps bool                 `json:"related_ops"`
-	ChainType  ChainType            `json:"chain_type"`
-	Payment    *ValidationOperation `json:"payment"`
-	Fee        *ValidationOperation `json:"fee"`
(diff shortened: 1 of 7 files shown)


asserter/asserter.go
```diff
@@ -46,10 +46,11 @@ type Asserter struct {
 // on the transaction. Fore more details please refer to
 // https://github.com/coinbase/rosetta-sdk-go/tree/master/asserter#readme
 type Validations struct {
-	Enabled   bool                 `json:"enabled"`
-	ChainType ChainType            `json:"chain_type"`
-	Payment   *ValidationOperation `json:"payment"`
-	Fee       *ValidationOperation `json:"fee"`
+	Enabled    bool                 `json:"enabled"`
+	RelatedOps bool                 `json:"related_ops"`
```
shrimalmadhur, line 50: Should we rename it something like `related_op_present` or `related_op_exists`? or something which sounds more like a boolean? just `RelatedOps` is not clear of what that param is.

shrimalmadhur:
You might want to run `make gen` locally to format and lint things.

shrimalmadhur (approved):

PR #356, opened 2021-11-19
fee should't have related op

shrimalmadhur:
Fixes # .

### Motivation
Fee op shouldn't have related_operation field

### Solution
added validation for the above scenario in the asserter.

### DIFF
--- a/asserter/errors.go
+++ b/asserter/errors.go
@@ -70,8 +70,8 @@ var (
 	ErrRelatedOperationIndexOutOfOrder = errors.New(
 		"related operation has index greater than operation",
 	)
-	ErrRelatedOperationIndexDuplicate     = errors.New("found duplicate related operation index")
-	ErrRelatedOperationMissing            = errors.New("related operations key is missing")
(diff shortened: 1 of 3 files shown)


asserter/errors.go
```diff
@@ -70,8 +70,11 @@ var (
 	ErrRelatedOperationIndexOutOfOrder = errors.New(
 		"related operation has index greater than operation",
 	)
-	ErrRelatedOperationIndexDuplicate     = errors.New("found duplicate related operation index")
-	ErrRelatedOperationMissing            = errors.New("related operations key is missing")
+	ErrRelatedOperationIndexDuplicate  = errors.New("found duplicate related operation index")
```
xiaying-peng, line 73: nit: align all `=` in the same vertical line..

xiaying-peng (approved):

asserter/errors.go
shrimalmadhur: formatting is correct. it's the github view messing around :)

PR #358, opened 2021-11-24
Fix: Use amount of 0 for currencies not listed

fkneeland-figure:
Fixes https://github.com/coinbase/rosetta-cli/issues/230

### Motivation

When using rosetta with cosmos-sdk based blockchains with coins minted after Genesis rosetta-cli tests error because of missing account balances. 

### Solution

To solve this issue, instead of returning an error when searching for a balance for a currency that doesn't have a value return a 0 amount.  This means that if rosetta looks for the amount of a currency at a block height before it has been minted it will simply return a 0 amount.

### DIFF
--- a/types/utils.go
+++ b/types/utils.go
@@ -305,19 +305,17 @@ func UnmarshalMap(metadata map[string]interface{}, output interface{}) error {
 func ExtractAmount(
 	balances []*Amount,
 	currency *Currency,
-) (*Amount, error) {
+) *Amount {
 	for _, b := range balances {
 		if Hash(b.Currency) != Hash(currency) {
 			continue
 		}
 
-		return b, nil
+		return b
 	}
 
-	return nil, fmt.Errorf(
-		"account balance response does not contain currency %s",
-		PrettyPrintStruct(currency),
-	)
+	// return a 0 amount if currency isn't found in balances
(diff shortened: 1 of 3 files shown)


fkneeland-figure:
This is ready for review.  Would @patrick-ogrady or @juliankoh be able to review this or tag someone on here who could?  Thanks!

shrimalmadhur:
> Fixes [coinbase/rosetta-cli#230](https://github.com/coinbase/rosetta-cli/issues/230)
> 
> ### Motivation
> When using rosetta with cosmos-sdk based blockchains with coins minted after Genesis rosetta-cli tests error because of missing account balances.
> 
> ### Solution
> To solve this issue, instead of returning an error when searching for a balance for a currency that doesn't have a value return a 0 amount. This means that if rosetta looks for the amount of a currency at a block height before it has been minted it will simply return a 0 amount.
> 
> ### Open questions

Hey @fkneeland-figure . Thanks for creating this. I will take a look at this.

xiaying-peng (dismissed):

fkneeland-figure:
@xiaying-peng thanks for the review!  I rebased and pushed up with a verified commit to meet merge requirements.

shrimalmadhur (approved):

PR #367, opened 2021-12-17
Bootstrap file memory issues

shrimalmadhur:
Fixes .
https://github.com/coinbase/rosetta-cli/issues/259 
https://github.com/coinbase/rosetta-cli/issues/260 

### Motivation
For motivation please refer #365 

### Solution
We commit transaction for bootstrap balances in chunks of 600 accounts.

### Open questions
I tested it locally with `MaxEntrySizePerTxn` to be lower/same than number of account and it worked. I still have not found a clean way to unit test that since that value comes from a const.

### DIFF
--- a/storage/modules/balance_storage.go
+++ b/storage/modules/balance_storage.go
@@ -1101,7 +1101,7 @@ func (b *BalanceStorage) BootstrapBalances(
 		// Commit transaction batch by batch rather than commit at one time.
 		// This helps reduce memory usage and improve running time when bootstrap_balances.json
 		// contains huge number of accounts.
-		if i%utils.MaxEntrySizePerTxn == 0 {
+		if i != 0 && i%utils.MaxEntrySizePerTxn == 0 {
 			if err := dbTransaction.Commit(ctx); err != nil {
 				return err
 			}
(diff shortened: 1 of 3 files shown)


storage/modules/balance_storage.go
```diff
@@ -1097,7 +1097,16 @@ func (b *BalanceStorage) BootstrapBalances(
 	dbTransaction := b.db.Transaction(ctx)
 	defer dbTransaction.Discard(ctx)
 
-	for _, balance := range balances {
+	for i, balance := range balances {
+		// Commit transaction batch by batch rather than commit at one time.
+		// This helps reduce memory usage and improve running time when bootstrap_balances.json
+		// contains huge number of accounts.
+		if i != 0 && i%utils.MaxEntrySizePerTxn == 0 {
```
xiaying-peng, line 1104: What about the last batch that is less than maxEntry?
ex: you have a list of 700, and maxEntry is 600.
Will we commit the [601, 700]?

shrimalmadhur: https://github.com/coinbase/rosetta-sdk-go/pull/367/files#diff-66c37bf21a824a6a6715142e0586224424d4c6cea795913fa549112ca51edc42R1142 - this line will commit the last tx. I need to see how I can write a unit test with smaller number to test it automatically but I tested it locally with maxEntry 2 with various cases.

rufsam:
Hi @shrimalmadhur, thanks for opening this PR. I was wondering if there is any update? Thanks alot!

xiaying-peng (approved):

PR #371, opened 2022-01-19
Improve parsing logic for `set_variable` action types

t-hale:
Fixes https://github.com/coinbase/rosetta-cli/issues/248

### Motivation
To improve the parsing logic for `set_variable` action types

### Solution
Before splitting the string to determine if the remaining line contains an arithmetic operator, we first check to see if the remaining line (sans `;` suffix) is valid JSON

### Open questions
Is there any possibility of arithmetic and JSON in the same line?

### DIFF
--- a/constructor/dsl/dsl_test.go
+++ b/constructor/dsl/dsl_test.go
@@ -330,6 +330,37 @@ func TestParse(t *testing.T) {
 			file:        "blah_blah.txt",
 			expectedErr: ErrIncorrectExtension,
 		},
+		"action: valid type": {
+			file: "action_valid.ros",
+			expectedWorkflows: []*job.Workflow{
+				{
+					Name:        string(job.CreateAccount),
+					Concurrency: job.ReservedWorkflowConcurrency,
+					Scenarios: []*job.Scenario{
+						{
+							Name: "create",
+							Actions: []*job.Action{
+								{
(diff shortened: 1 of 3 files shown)


shrimalmadhur:
> Is there any possibility of arithmetic and JSON in the same line?
Can you give me an example of what it will look like. I did not get this case?

t-hale:
> > Is there any possibility of arithmetic and JSON in the same line?
> 
> @t-hale Can you give me an example of what it will look like. I did not get this case?

I take it by your confusion that it's not possible.  I'm pretty new to the Rosetta DSL, so I wasn't sure if there was a possibility of these two cases `math` and `set_variable` occurring on the same line (this case would make the new logic I added in this PR invalid)

t-hale:
@shrimalmadhur bumping this one back up

shrimalmadhur:
@t-hale sorry about this delay. I think there's not possibility of JSON and math op in same line. so this should be good. I see some build failures - Possibly due to stale code. Could you please rebase? Thank you and again apologies for the delays.

t-hale:
> @t-hale sorry about this delay. I think there's not possibility of JSON and math op in same line. so this should be good. I see some build failures - Possibly due to stale code. Could you please rebase? Thank you and again apologies for the delays.

No worries - I've gone ahead and rebased

PR #376, opened 2022-02-01
Add Pallas support

irisZhangCB:
Fixes # .

### Motivation

Currently assets with Pallas curve cannot use rosetta-cli for validation because we don't support it in rosetta-sdk-go.
### Solution

Add Pallas support


keys/signer_edwards25519_test.go
```diff
@@ -78,7 +78,7 @@ func mockSignature(
 	payload := &types.SigningPayload{
 		AccountIdentifier: &types.AccountIdentifier{Address: "test"},
 		Bytes:             msg,
-		SignatureType:     types.Ed25519,
```
shrimalmadhur, line 81: Is this an intended change? I am wondering how this hardcoded change worked before?

keys/signer_pallas_test.go
```diff
@@ -0,0 +1,136 @@
+// Copyright 2020 Coinbase, Inc.
```
shrimalmadhur, line 1: can you change this to 2022?

keys/signer_pallas.go
```diff
@@ -0,0 +1,126 @@
+// Copyright 2020 Coinbase, Inc.
```
shrimalmadhur, line 1: 2022

keys/signer_edwards25519_test.go
shrimalmadhur: oh I see you refactored at the end.

shrimalmadhur (approved):

PR #385, opened 2022-02-22
Pallas support ImportPrivateKey

irisZhangCB:
Fixes # .

### Motivation

- ImportPrivateKey didn't add Pallas curve type. 
- There's discrepancy between how kryptology and rosetta process payload. Added functions to concert payload to work with kryptology


keys/signer_pallas.go
```diff
@@ -110,17 +113,110 @@ func (s *SignerPallas) Verify(signature *types.Signature) error {
...
+			return nil, err
+		}
+	}
+
+	nonce, err := strconv.ParseUint(p.Nonce, 10, 32)
+	if err != nil {
+		return nil, err
+	}
+
+	validUntil := uint64(0)
```
shrimalmadhur, line 191: Why do we have to define it `uint64` when we end up casting it into `uint32` ?

shrimalmadhur: ah `ParseUint` returns `uint64`

keys/signer_pallas_test.go
```diff
@@ -28,31 +27,27 @@ var keypair *KeyPair
 var txnBytes []byte
 
 func init() {
-	keypair, _ = GenerateKeypair(types.Pallas)
+	keypair, _ = ImportPrivateKey(
```
shrimalmadhur, line 30: Can we add more test cases for covering edge cases since there are lot of `if` loops while executing `constructTransaction`

shrimalmadhur (approved):

PR #387, opened 2022-03-05
implement Signer interface for bls12381

freddiecoleman:
Fixes # .

### Motivation

This allows Rosetta users to sign and verify payloads using a `bls12381` curve type.

### Solution

Provide an implementation of the Signer interface for `bls12381`

### Open questions

- It looks like the build/pipeline pulls a JSON file containing the specification and the `bls12381` curve type will not currently be in there. I believe adding the necessary support will require a further PR somewhere else - this should be quite simple?


shrimalmadhur:
> ### Motivation
> This allows Rosetta users to sign and verify payloads using a `bls12381` curve type.
> 
> It supports both basic and aug with prefix by passing in the relevant signature type.
> 
> In order to sign using aug you will need to prepend the relevant public key to the payload that is to be signed as the exported interface does not allow an additional argument for the public key.
> 
> ### Solution
> Provide an implementation of the Signer interface for `bls12381`
> 
> ### Open questions
> * It looks like the build/pipeline pulls a JSON file containing the specification and the `bls12381` curve type will not currently be in there. I believe adding the necessary support will require a further PR somewhere else - this should be quite simple?

Hello @freddiecoleman - you need to make specifications changes here - https://github.com/coinbase/rosetta-specifications
You can use https://github.com/coinbase/rosetta-specifications/pull/93 as sample.

PR #391, opened 2022-03-21
ROSE-230: Use integer to compare amount sign

shiatcb:
### Motivation
https://jira.coinbase-corp.com/browse/ROSE-230

### Solution
Currently we use amount.Value == "0" to check if an amount matches zero. This caused an issue in ChainIO as we may have "-0" as input. To resolve the issue, we need to convert string to integer and then compare.

### DIFF
--- a/parser/match_operations.go
+++ b/parser/match_operations.go
@@ -67,11 +67,11 @@ func (s AmountSign) Match(amount *types.Amount) bool {
 		return true
 	}
 
-	if s == PositiveOrZeroAmountSign && (numeric.Sign() == 1 || amount.Value == "0") {
+	if s == PositiveOrZeroAmountSign && (numeric.Sign() == 1 || len(numeric.Bits()) == 0) {
 		return true
 	}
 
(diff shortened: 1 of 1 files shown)


parser/match_operations.go
```diff
@@ -67,11 +67,11 @@ func (s AmountSign) Match(amount *types.Amount) bool {
 		return true
 	}
 
-	if s == PositiveOrZeroAmountSign && (numeric.Sign() == 1 || amount.Value == "0") {
+	if s == PositiveOrZeroAmountSign && (numeric.Sign() == 1 || numeric.Cmp(big.NewInt(0)) == 0) {
```
xiaying-peng, line 70: We can take a look at https://stackoverflow.com/questions/64257065/is-there-another-way-of-testing-if-a-big-int-is-0
bits compare is proven to be 20 times faster than comparing it to another big.Int representing 0

shiatcb: Great comment @xiaying-peng ! Checked and updated PR.

xiaying-peng (approved):

PR #395, opened 2022-04-04
minor optimization

raghavapamula:
Fixes # .


syncer/syncer.go
```diff
@@ -646,7 +646,7 @@ func (s *Syncer) Sync(
 			log.Printf("Syncing %d\n", s.nextIndex)
 		}
 
-		err = s.syncRange(ctx, rangeEnd)
+		err = s.syncRange(ctx, rangeEnd, endIndex-s.nextIndex+1)
```
shrimalmadhur, line 649: Can you explain this optimization. In both the cases we are executing `endIndex-s.nextIndex+1` once. Are saying the optimization is creating of local variable `blocksToSync`? And it would perform better if passed through function arugment?

raghavapamula: yeah, blocksToSync both has to be allocated and then later de-allocated by the garbage collector. Not a huge performance hit, but based on the benchmarks I ran, it should take a few minutes off syncing for a large blockchain such as Terra Luna.

Happy to walk through the benchmarks I ran as well

shrimalmadhur: Do you see consistently better performance with this change? As in if you run the same experiment let's say 10 times, do you see this improvement 10 times?

raghavapamula: I took another look at the code and ran the experiment a few more times, i now think this doesn't really improve performance. I'm gonna close the PR, thanks for looking

shrimalmadhur: yea. I suspected so. local variables doesn't add to perf that much. Thanks for investigating.

PR #397, opened 2022-04-12
Add validation for negative fee amount

raghavapamula:
Fixes # https://jira.coinbase-corp.com/browse/ROSE-90

### Motivation
Operations of type fee must have negative value

### Solution
Add validation for this behaviour


asserter/block.go
```diff
@@ -320,7 +320,18 @@ func (a *Asserter) Operations( // nolint:gocognit
 					)
 				}
 
-				val, _ := new(big.Int).SetString(op.Amount.Value, 10)
+				val, err := types.BigInt(op.Amount.Value)
+				if err != nil {
+					return err
+				}
+				if val.Sign() != -1 {
```
xiaying-peng, line 327: Let's add test coverage

raghavapamula: done

asserter/block_test.go
```diff
@@ -513,6 +521,52 @@ func TestOperationsValidations(t *testing.T) {
 			construction:       false,
 			err:                ErrRelatedOperationInFeeNotAllowed,
 		},
+
+		"fee amount is non-negative": {
```
xiaying-peng, line 525: this is the happy path, we want error case test as well

xiaying-peng (dismissed):

xiaying-peng (approved):

PR #402, opened 2022-04-26
Replace CircleCI with gh actions

irisZhangCB:
Fixes # .

### Motivation

Replace CircleCI with gh actions


.github/workflows/ci.yml
```diff
@@ -0,0 +1,80 @@
+name: CI
```
shrimalmadhur, line 1: `check-gen` is missing in this.

shrimalmadhur: I guess `setup` is replaced by `build` - so just `check-gen` then

shrimalmadhur (approved):

PR #414, opened 2022-06-07
add sub account optional flag

### LINKED ISSUE #403: Support parse sub account as optional

As of the parser of SDK, the sub account is either required or not.  There is no option to make subaccount field as optional.

A flag to parse account with subaccount string as optional.

**Additional context**

shrimalmadhur:
Fixes #403  .

### Motivation
In some rosetta implementation `sub account` field is optional. This means we can't predict if it will be there or not. The current matcher logic requires it to be either required or not required. 

### Solution
Added a flag called `SubAccountOptional` which if set to `true` will ignore the subaccount matching and if false, behaves exactly as before

### DIFF
--- a/parser/match_operations.go
+++ b/parser/match_operations.go
@@ -105,7 +105,11 @@ type MetadataDescription struct {
 
 // AccountDescription is used to describe a *types.AccountIdentifier.
 type AccountDescription struct {
-	Exists                 bool
+	Exists bool
+
+	// SubAccountOptional If this is true then SubAccountExists, SubAccountAddress,
+	// SubAccountMetadataKeys matching is ignored
+	SubAccountOptional     bool
 	SubAccountExists       bool
 	SubAccountAddress      string
 	SubAccountMetadataKeys []*MetadataDescription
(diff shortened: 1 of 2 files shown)


parser/match_operations_test.go
```diff
@@ -1408,6 +1408,132 @@ func TestMatchOperations(t *testing.T) {
 			matches: nil,
 			err:     true,
 		},
+		"nil amount ops (sub account optional - when sub-account exist)": {
+			operations: []*types.Operation{
+				{
+					Account: &types.AccountIdentifier{
```
ningcb, line 1414: Can we add a group of operations in which one with sub account, another does not?

parser/match_operations.go
```diff
@@ -213,6 +217,10 @@ func accountMatch(req *AccountDescription, account *types.AccountIdentifier) err
 		return nil
 	}
 
+	if req.SubAccountOptional {
```
ningcb, line 220: if SubAccountOptional true, should it check SubAccountAddress on the account with subaccount?

shrimalmadhur: I am thinking if it is optional, do we care about what is inside that?

parser/match_operations_test.go
shrimalmadhur: updated this.

ningcb (approved):

cb-heimdall:
Review Error for ningcb @ 2022-06-08 20:38:36 UTC
User must have write permissions to review

irisZhangCB (approved):

PR #418, opened 2022-06-27
Destination Tag Support

GeekArthur:
Signed-off-by: Jingfu Wang <jingfu.wang@coinbase.com>

### Motivation

This can enable rosetta validation (i.e. rosetta cli) support validation of rosetta implementation with destination tag technology 

### Solution

We need to broadcast transaction with destination tag then validate the destination tag from the on-chain transaction


utils/utils.go
```diff
@@ -447,6 +447,38 @@ func CurrencyBalance(
 	return liveAmount, liveBlock, nil
 }
 
+// CurrencyBalanceBatch returns the balance batch of an account
+// for all currencies at a particular height.
+// It is up to the caller to determine if the retrieved
+// block has the expected hash for the requested index.
+func CurrencyBalanceBatch(
```
xiaying-peng, line 454: Why this change is related to dest tag?

GeekArthur: This is related the support of getting all balances of the pre-funded account when currency is nil, the reason we add this is because we use STX to test destination tag and STX rosetta implementation `/account/balance` only support nil currency, which is compliant our rosetta spec. However, out rosetta cli/sdk has the limitation of doesn't support nil currency, add this for both testing STX and scaling rosetta cli/sdk

xiaying-peng: Maybe rename this method to `AllCurrencyBalance`?

GeekArthur: Sure, will update

utils/utils.go
```diff
@@ -474,26 +506,46 @@ func GetAccountBalances(
...
-
-		if err != nil {
-			return nil, err
-		}
-
-		accountBalance := &AccountBalance{
-			Account: balanceRequest.Account,
-			Amount:  amount,
-			Block:   block,
+		if balanceRequest.Currency != nil {
```
shrimalmadhur, line 509: Can you add unit tests for this change?

GeekArthur: Yup, will add them

PR #423, opened 2022-07-06
Add hooks to allow overriding Helper field in Syncer

### LINKED ISSUE #422: Handling account aliases on the Concordium blockchain (rosetta-cli)

On the [Concordium](https://www.concordium.com/) blockchain we have the concept of account aliases, i.e. only the first 29 (of 53) bits of an account address are significant. This confuses `rosetta-cli check:data` as it has no way of knowing if two addresses actually reference the same account. If an operation was performed for some address, it will observe that the balance of the "account" of the aliases changed for no reason.

I've been trying to find a way to patch `rosetta-cli` (via `rosetta-sdk-go`) to translate all addresses into the "canonical" one. I've gotten it working by adding the following abomination of a hack into `Syncer.processBlock`:

```
@@ -187,6 +229,11 @@ func (s *Syncer) processBlock(
        }
 
        block := br.block
+       for i := range block.Transactions {
+               for j := range block.Transactions[i].Operations {
+                       canonicalizeAccountAddr(block.Transactions[i].Operations[j].Account)
+               }
+       }
        err = s.handler.BlockAdded(ctx, block)
        if err != nil {
                return err

```

where `canonicalizeAccountAddr` does a lookup using the Concordium SDK and replaces the address in-place.

I've not been able to find a _proper_ place to implement the behavior as `Syncer`, `StatefulSyncer`, `Fetcher`, etc. are all structs and not interfaces. `Helper` is an interface type but it's not being constructed in a factory that can be overriden.

I'm not sure if monkey patching the block result is the correct solution, but it's the only one I've gotten to work (also tried `BalanceStorage.AddingBlock` and `Parser.BalanceChanges` but those seemed insufficient).

Patching all accounts in our Rosetta implementation. That would defeat the purpose of having aliases and also be prohibitively expensive.

**Additional context**

Any ideas of the correct way to implement this are greatly appreciated. I'll be happy to propose a PR with the necessary changes to for example support overriding the behavior of `Syncer`.

bisgardo:
Fixes #422.

### Motivation

For `rosetta-cli check:data` to do accounting correctly on the Concordium blockchain, it needs to rewrite account addresses into the same "normalized" alias.

### Solution

Add (stateful)syncer options for overriding `syncer.Option`s passed to `syncer.New` from outside the `statefulsyncer.New`.

Our internal fork of `rosetta-cli` is able to utilize this in https://github.com/Concordium/rosetta-cli/pull/1/files.

### Open questions

There might be better ways to solve the underlying problem than patching block results, but I haven't been able to find one that works as reliably. So even if this is a bit brute-force'ish it's good enough for me.

### DIFF
--- a/statefulsyncer/configuration.go
+++ b/statefulsyncer/configuration.go
@@ -16,6 +16,8 @@ package statefulsyncer
 
 import (
 	"time"
+
+	"github.com/coinbase/rosetta-sdk-go/syncer"
 )
 
 // Option is used to overwrite default values in
@@ -64,5 +66,11 @@ func WithPruneSleepTime(sleepTime int) Option {
 func WithSeenConcurrency(concurrency int64) Option {
 	return func(s *StatefulSyncer) {
 		s.seenSemaphoreSize = concurrency
+	}
+}
+
+func WithExtraSyncerOpts(os ...syncer.Option) Option {
+	return func(s *StatefulSyncer) {
+		s.extraSyncerOpts = os
 	}
 }
--- a/statefulsyncer/stateful_syncer.go
+++ b/statefulsyncer/stateful_syncer.go
@@ -72,6 +72,8 @@ type StatefulSyncer struct {
 	// BlockSeen occur concurrently.
 	seenSemaphore     *semaphore.Weighted
 	seenSemaphoreSize int64
+
+	extraSyncerOpts []syncer.Option
 }
 
 // Logger is used by the statefulsyncer to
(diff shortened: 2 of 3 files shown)


statefulsyncer/configuration.go
```diff
@@ -66,3 +68,9 @@ func WithSeenConcurrency(concurrency int64) Option {
 		s.seenSemaphoreSize = concurrency
 	}
 }
+
+func WithExtraSyncerOpts(os ...syncer.Option) Option {
+	return func(s *StatefulSyncer) {
+		s.extraSyncerOpts = os
+	}
+}
```
bisgardo, line 76: A more flexible solution would be to pass a mapping function like in `WithCustomHelper` below. But that's probably unnecessarily complex.

bisgardo:
Is anyone ever going to take a look at this? @jingweicb @shrimalmadhur @GeekArthur

jingweicb:
> Is anyone ever going to take a look at this? @jingweicb @shrimalmadhur @GeekArthur

@bisgardo sorry for making you waiting, will circle back later

PR #428, opened 2022-07-26
Update .gitignore to exclude .vscode

xiaying-peng:
Fixes # .

### Motivation
Improve gitignore

### Solution
Update .gitignore to exclude .vscode


GeekArthur:
Should we add this back? I think we already discussed this in https://github.com/coinbase/rosetta-cli/pull/333#discussion_r909130827, @shrimalmadhur

PR #432, opened 2022-08-12
Fix eth vevrsion alert

jingweicb:
Fixes # .

### Solution
import the ethereum repo

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -36,6 +36,7 @@ require (
 	github.com/golang/protobuf v1.5.2 // indirect
 	github.com/golang/snappy v0.0.4 // indirect
 	github.com/klauspost/compress v1.12.3 // indirect
+	github.com/kr/pretty v0.2.1 // indirect
 	github.com/mattn/go-colorable v0.1.9 // indirect
 	github.com/mattn/go-isatty v0.0.14 // indirect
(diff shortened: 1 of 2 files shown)


jingweicb:
the previous breaking is because https://github.com/coinbase/rosetta-sdk-go/runs/7793099183?check_suite_focus=true 
what I do is 
1 run go get github.com/ethereum/go-ethereum 
2 go mod tiy 
3 make check-gen
@shrimalmadhur

go.mod
```diff
@@ -36,6 +36,7 @@ require (
 	github.com/golang/protobuf v1.5.2 // indirect
 	github.com/golang/snappy v0.0.4 // indirect
 	github.com/klauspost/compress v1.12.3 // indirect
+	github.com/kr/pretty v0.2.1 // indirect
```
shrimalmadhur, line 39: wait this is not eth?

jingweicb: not, but I see this add into here when run  go get github.com/ethereum/go-ethereum

shrimalmadhur: also the latest is `v1.10.21` I think the problem is latest has some issues. check out line 12 - we use latest. so we can't do anything for now

PR #436, opened 2022-08-23
Error handling improvement for asserter package

GeekArthur:
Signed-off-by: Jingfu Wang <jingfu.wang@coinbase.com>

Error handling improvement for asserter package

### DIFF
--- a/asserter/construction_test.go
+++ b/asserter/construction_test.go
@@ -201,6 +201,14 @@ func TestConstructionDeriveResponse(t *testing.T) {
 		"nil response": {
 			err: ErrConstructionDeriveResponseIsNil,
 		},
+		"empty address": {
+			response: &types.ConstructionDeriveResponse{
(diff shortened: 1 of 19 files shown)


jingweicb (approved):

jingweicb:
@GeekArthur this is a huge change! Great work!

asserter/construction_test.go
```diff
@@ -259,40 +250,6 @@ func TestConstructionParseResponse(t *testing.T) {
 			signed: true,
 			err:    nil,
 		},
-		"duplicate signer": {
```
shrimalmadhur, line 262: why are these tests removed?

GeekArthur: @shrimalmadhur My original thinking was we can remove it the error is not generated from this package, but it also doesn't hurt to include it for better testing coverage, I added it back

shrimalmadhur (approved):

PR #457, opened 2022-12-05
Pallas signer stake delegation

ylecornec:
### Motivation
The following PR updates the pallas signer used for mina.
- It adds support for stake delegation transactions, in order to test the mina rosetta implementation.
- It fixes the default value for the `validUntil` field of payment transaction (from `0` to `max_uint32`).

This was tested by running the construction tests on the mina rosetta server with both payment and delegation transactions.

#### Delegation transactions
The `Transaction` type could already handle various types of transaction and distinguish between them using the `Tag` field.
`{false, false, false}` is the first case (payments) and `{false,false,true}` the second one (stake delegation).
However the parsing step in `signer_pallas.go` could only handle `Payment` transaction so far so I added the `StakeDelelegation` case.

#### The valid_until field

If the `valid_until` field is not specified in the preprocess step, the rosetta server uses the greatest `uint32` as a default value. The pallas signer of `rosetta-sdk-go` was defaulting to zero when building the signing payload which resulted in an `invalid signature` error. The greatest `uint32` is now used in both cases.

### DIFF
--- a/keys/errors.go
+++ b/keys/errors.go
@@ -44,7 +44,7 @@ var (
 	)
 	ErrVerifyFailed = errors.New("verify: verify returned false")
 
-	ErrPaymentNotFound = errors.New("payment not found in signingPayload")
+	ErrEmptyTransaction = errors.New("payment or stakeDelegation field not found in signingPayload")
 )
 
 // Err takes an error as an argument and returns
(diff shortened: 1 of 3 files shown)


ylecornec:
The tests should now be fixed (`make test` succeeds locally) and I added `delegation` versions of existing `payment` tests.

ylecornec:
Closing in favor of #464

PR #458, opened 2022-12-06
chore: bump btcsuite/btcd

julienrbrt:
The version used should not be used anymore: https://github.com/advisories/GHSA-2chg-86hq-7w38.
Projects checking their dependencies with https://github.com/actions/dependency-review-action are getting warnings (e.g. https://github.com/cosmos/cosmos-sdk/actions/runs/3629277527/jobs/6121284459).


go.mod
```diff
@@ -25,11 +25,13 @@ require (
 
 require (
 	filippo.io/edwards25519 v1.0.0-rc.1 // indirect
+	github.com/btcsuite/btcd v0.21.0-beta.0.20201114000516-e9c7a5ac6401 // indirect
```
julienrbrt, line 28: Seems like it will be hard to remove this looking at go mod graph.

PR #485, opened 2023-05-31
update retry logic

jingweicb:
Fixes # .

We met this problem in most of recent validation by using internal tools:
Many errors's use default logic which is not retriable, however, due to network problems we saw many timeout should be retriable(we found this in /block, /account/balance, and /network/options)

This pr adds the `force_retry` to the `tryagain` function which solves it

### Motivation

do not merge it or change to ready for review at this moment

### DIFF
--- a/fetcher/utils.go
+++ b/fetcher/utils.go
@@ -81,8 +81,8 @@ func transientError(err error) bool {
 
 // tryAgain handles a backoff and prints error messages depending
 // on the fetchMsg.
-func tryAgain(fetchMsg string, thisBackoff *Backoff, err *Error) *Error {
-	if !err.Retry {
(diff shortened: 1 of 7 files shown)


fetcher/utils.go
```diff
@@ -81,8 +81,8 @@ func transientError(err error) bool {
 
 // tryAgain handles a backoff and prints error messages depending
 // on the fetchMsg.
-func tryAgain(fetchMsg string, thisBackoff *Backoff, err *Error) *Error {
-	if !err.Retry {
+func tryAgain(fetchMsg string, forceRetry bool, thisBackoff *Backoff, err *Error) *Error {
+	if !err.Retry && !forceRetry {
```
jingweicb, line 85: another way is to remove `err.Retry` judgment  to make all the fetcher errors retriable
as this is for fetcher, it will not influence our accuracy

PR #489, opened 2023-07-22
Guarantee the order of transactions in the block endpoint

wei-wang-cb:
### Motivation

This commit guarantees the order of transactions in the block endpoint. This can benefit clients in a certain way. For example, some Rosetta client requires transaction kept in order so that they can identify the coinbase (lower-case c) transaction in one block.

### Solution

It does so by creating a map of fetched transactions and using that map to return the transactions in the original order they had in the block.

- Added a new struct `FetchedTransaction` to represent fetched transactions
- Updated `fetchChannelTransactions` to use the new `FetchedTransaction` struct
- Updated `UnsafeTransactions` to use the new `FetchedTransaction` struct and create a map of fetched transactions
- Modified the return statement in `UnsafeTransactions` to use the fetched transactions from the map

This PR also improves some security issues raised by salus check.

### DIFF
--- a/client/api_account.go
+++ b/client/api_account.go
@@ -19,6 +19,7 @@ package client
 import (
 	_context "context"
 	"fmt"
+	"io"
 	_ioutil "io/ioutil"
 	_nethttp "net/http"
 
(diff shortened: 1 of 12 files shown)


GeekArthur:
two generic comments (i will leave them in the PR as well):
1. please run rosetta validation for Dogecoin and at least one non-Dogecoin asset to ensure there are no regressions as rosetta-sdk-go is a common library that is used by many Rosetta assets and Rosetta validation itself
2. please track the performance difference as seems like we add additional linear layer for sorting so the time complexity can change from O(n) to O(c * n)

wei-wang-cb:
Hi @GeekArthur , I've done rosetta validation (check-data) for two asset and they all passed. Could you help me merge this PR?

GeekArthur (approved):

xiaying-peng:
are the api_xyz.go changes generated by `codegen.sh`?
We should add minimal test coverage on these kinds of changes.
cc @GeekArthur

PR #502, opened 2024-01-31
Add a new asserter client with ignoring rosetta spec validation.

jacquescaocb:
Fixes # .

### Motivation

in somecases, rosetta spec validation is not helpful, and it causes some operation work, so now we add a new client which can ignore rosetta spec validation.

### DIFF
--- a/asserter/asserter.go
+++ b/asserter/asserter.go
@@ -1,4 +1,4 @@
-// Copyright 2020 Coinbase, Inc.
+// Copyright 2024 Coinbase, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/asserter/asserter.go
+++ b/asserter/asserter.go
@@ -294,9 +294,9 @@ func NewClientWithOptions(
 	return asserter, nil
 }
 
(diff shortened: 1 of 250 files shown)


asserter/asserter.go
```diff
@@ -292,6 +294,47 @@ func NewClientWithOptions(
...
+	asserter := &Asserter{
+		network:      network,
+		genesisBlock: genesisBlockIdentifier,
+		validations: &Validations{
+			Enabled: false,
+		},
+		ignoreRosettaSpecValidation: true,
+	}
+
+	asserter.operationStatusMap = map[string]bool{}
```
xiaying-peng, line 328: let's move this to a const list in validation.go

asserter/operation_status.go
```diff
@@ -0,0 +1,34 @@
+// Copyright 2020 Coinbase, Inc.
```
xiaying-peng, line 1: 2024？？？

jacquescaocb: from what I saw, we have never changed copyright in asserter last 4 years. only those files in types, client, and server directories have been changed every year. 

but anyway, I just changed all files which have copyright 2020.

xiaying-peng (approved):

PR #507, opened 2024-10-17
feat(header forwarding): add HeaderForwarder and plumbing

potterbm-cb:
### Motivation

In order to support some common use cases, Mesh servers need the ability to pass request headers through to the native nodes, and to return headers from native node requests back to the caller.

Use cases:
* **passing a client ID header to the nodes** — this client ID shouldn't originate with the Mesh server since it may have many different callers, so it should be passed through to the native node requests
* **sticky session cookies** — to support sticky sessions on native nodes, AWS uses cookies. Cookies are managed by the `Cookie` and `Set-Cookie` headers, which will need to be passed from the caller to the native node, and returned from the native node ALB to the caller

### Solution

The `HeaderForwarder` type extracts headers from potentially many native node requests and adds them to a Mesh response. HeaderForwarder can be injected into native node network requests as an `http.RoundTripper`, and can be applied to Mesh servers as middleware

A few key features:
* implements `http.RoundTripper` so that it can be used to create an `http.Client` or other client
* implements a common middleware pattern: `func(http.Handler) http.Handler`
* assigns every request a unique ID (which is used to associate headers from native node responses back to Mesh request/responses)
* allows forwarding only specific headers

### Open questions

1. Currently there is no proof-of-concept for non-geth chains, so it remains to be seen if this architecture will work well for cosmos chains or other.

### DIFF
--- a/headerforwarder/forwarder.go
+++ b/headerforwarder/forwarder.go
@@ -1,0 +1,123 @@
+package headerforwarder
+
+import (
+	"fmt"
+	"net/http"
+)
+
+// HeaderExtractingTransport is a utility to help a rosetta server forward headers to and from
+// native node requests. It implements several interfaces to achieve that:
+//   - http.RoundTripper: this can be used to create an http Client that will automatically save headers
+//     if necessary
+//
+// the headers can be requested later.
+//
+// TODO: this needs to expire entries after a certain amount of time
+type HeaderForwarder struct {
+	requestHeaders     map[string]http.Header
+	interestingHeaders []string
+	actualTransport    http.RoundTripper
+}
+
+func NewHeaderForwarder(interestingHeaders []string, transport http.RoundTripper) *HeaderForwarder {
+	return &HeaderForwarder{
+		requestHeaders:     make(map[string]http.Header),
+		interestingHeaders: interestingHeaders,
+		actualTransport:    transport,
+	}
+}
+
+// RoundTrip implements http.RoundTripper and will be used to construct an http Client which
+// saves the native node response headers if necessary.
+func (hf *HeaderForwarder) RoundTrip(req *http.Request) (*http.Response, error) {
+	resp, err := hf.actualTransport.RoundTrip(req)
+
+	if err == nil && hf.shouldRememberHeaders(req, resp) {
+		hf.rememberHeaders(req, resp)
+	} else {
+		fmt.Println("not remembering headers", resp.Header)
+	}
+
+	return resp, err
+}
+
+// shouldRememberHeaders is called to determine if response headers should be remembered for a given request.
+// Response headers will only be remembered if the request does not contain all of the interesting
+// headers and the response contains at least one of the interesting headers.
+//
+// It should be noted that the request and response here are for a request to the native node,
+// not a request to the Rosetta server.
+func (hf *HeaderForwarder) shouldRememberHeaders(req *http.Request, resp *http.Response) bool {
+	requestHasAllHeaders := true
+	responseHasSomeHeaders := false
+
+	for _, interestingHeader := range hf.interestingHeaders {
+		_, requestHasHeader := req.Header[http.CanonicalHeaderKey(interestingHeader)]
+		_, responseHasHeader := resp.Header[http.CanonicalHeaderKey(interestingHeader)]
+
+		if !requestHasHeader {
+			requestHasAllHeaders = false
+		}
+
+		if responseHasHeader {
+			responseHasSomeHeaders = true
+		}
+	}
+
+	// only remember headers if the request does not contain all of the interesting headers and the
+	// response contains at least one
+
+	return !requestHasAllHeaders && responseHasSomeHeaders
+}
+
+// rememberHeaders is called to save the native node response headers. The request object
+// here is a native node request (constructed by go-ethereum for geth-based rosetta implementations).
+// The response object is a native node response.
+//
+// TODO: only remember interesting headers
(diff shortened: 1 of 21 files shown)


cb-heimdall:
<h1>🟡 Heimdall Review Status</h1>
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<summary>Denominator calculation</summary>
<table>
<tr>
<td>
<details>
<summary>Show calculation</summary>
<table>
<tr>
<td><code>1</code> if user is bot</td>
<td>0</td>
</tr>
<tr>
<td><code>1</code> if user is external</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>.codeflow.yml</code>
</td>
<td>1</td>
</tr>
<tr>
<td>
Additional review requirements
<details>
<summary>Show calculation</summary>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td><code>1</code> if commit is unverified</td>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

headerforwarder/forwarder.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+// HeaderExtractingTransport is a utility to help a rosetta server forward headers to and from
+// native node requests. It implements several interfaces to achieve that:
+//   - http.RoundTripper: this can be used to create an http Client that will automatically save headers
+//     if necessary
+//
+// the headers can be requested later.
+//
+// TODO: this needs to expire entries after a certain amount of time
```
marklandgrebe-cb, line 15: Instead of expiring after a certain amount of time, can we delete an entry from requestHeaders after it is looked up and added to the final rosetta response's headers?

headerforwarder/forwarder.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// rememberHeaders is called to save the native node response headers. The request object
+// here is a native node request (constructed by go-ethereum for geth-based rosetta implementations).
+// The response object is a native node response.
+//
+// TODO: only remember interesting headers
+func (hf *HeaderForwarder) rememberHeaders(req *http.Request, resp *http.Response) {
+	ctx := req.Context()
+	// rosettaRequestID := services.osettaIdFromContext(ctx)
```
marklandgrebe-cb, line 80: nit: remove

headerforwarder/forwarder.go
```diff
@@ -0,0 +1,123 @@
...
+		hf.rememberHeaders(req, resp)
+	} else {
+		fmt.Println("not remembering headers", resp.Header)
+	}
+
+	return resp, err
+}
+
+// shouldRememberHeaders is called to determine if response headers should be remembered for a given request.
+// Response headers will only be remembered if the request does not contain all of the interesting
```
marklandgrebe-cb, line 45: What if a sticky session cookie is provided in the request but it expires and a new sticky session cookie comes in the response?

headerforwarder/forwarder.go
```diff
@@ -0,0 +1,123 @@
...
+// rememberHeaders is called to save the native node response headers. The request object
+// here is a native node request (constructed by go-ethereum for geth-based rosetta implementations).
+// The response object is a native node response.
+//
+// TODO: only remember interesting headers
+func (hf *HeaderForwarder) rememberHeaders(req *http.Request, resp *http.Response) {
+	ctx := req.Context()
+	// rosettaRequestID := services.osettaIdFromContext(ctx)
+	rosettaRequestID := RosettaIDFromContext(ctx)
+	fmt.Printf("remembering headers for request id %s: %v\n", rosettaRequestID, resp.Header)
```
marklandgrebe-cb, line 82: Likely too noisy and should be removed

jingweicb:
Hi @potterbm-cb , some CI checks failed, can you try to fix them, same for the other pr.

headerforwarder/forwarder.go
potterbm-cb: technically since the `Cookie` and `Set-Cookie` header are different, headers for requests with cookies will always be forwarded.

If that's eventually not what we want we'd have to write custom logic to associate `Cookie` and `Set-Cookie` together for request/response (and storing/accessing) respectively

headerforwarder/forwarder.go
potterbm-cb: This is a great idea (and one I meant to add but forgot). 

Eventually we should probably do both just because headers that are never accessed would stick around forever, but deleting entries as we use them will mean we can release this without expiration probably

marklandgrebe-cb (dismissed):

marklandgrebe-cb (approved):
lgtm, we should work on covering this repo with unit tests

jingweicb (approved):

potterbm-cb:
closed and reopened PR to try to get the CI unstuck ("Minimum required reviews" wasn't registering the reviews)

xiaying-peng (approved):

PR #538, opened 2026-04-06
fix(types): encode nil slices as empty arrays in Allow type

### LINKED ISSUE #62: Encode nil slices as empty arrays
By default, the `json` package marshals zero-valued (nil) slices as `null`, whereas merely *empty* slices are marshalled as `[]`:

```go
var s []int
json.Marshal(s) // -> null
s = []int{}
json.Marshal(s) // -> []
```

If the JSON is later consumed by a Go client, this generally isn't an issue, since both `null` and `[]` will be decoded as a slice with no elements, and Go code typically doesn't distinguish between nil and empty slices. However, clients in other languages may choke on `null`, e.g.:
```js
let arr = request(endpoint)
let len = arr.length // invalid if arr === null
```

To avoid this, zero-valued slices should be replaced with empty slices prior to encoding. There are a few ways to do this, but unfortunately none of them are trivial. Perhaps the best approach for Rosetta would be to use a JSON codegen tool such as `easyjson`, which supports a `NilSliceAsEmpty` option. This would also improve encoding/decoding performance.

Kubudak90:
Fixes #62

## Problem
By default, Go's json package marshals nil slices as null, whereas empty slices are marshaled as []. This causes issues for clients in other languages (e.g., JavaScript) that may choke on null when expecting an array.

## Solution
Added a custom MarshalJSON method to the Allow type that converts nil slices to empty slices before marshaling. This ensures consistent encoding as [] regardless of whether the slice is nil or empty.

## Changes
- Added MarshalJSON method to Allow type
- Nil slices for OperationStatuses, OperationTypes, Errors, CallMethods, and BalanceExemptions are now encoded as [] instead of null
- Added test file nil_slice_test.go with tests for nil and empty slice marshaling

## Testing
The fix ensures:
- nil slices → [] in JSON
- Empty slices → [] in JSON (unchanged behavior)
- Non-empty slices → original values in JSON (unchanged behavior)

This is a backward-compatible change that only affects the JSON representation of nil slices, making the API more robust for cross-language compatibility.

### DIFF
--- a/types/allow.go
+++ b/types/allow.go
@@ -16,6 +16,8 @@
 
 package types
 
+import "encoding/json"
+
 // Allow Allow specifies supported Operation status, Operation types, and all possible error
 // statuses. This Allow object is used by clients to validate the correctness of a Rosetta Server
 // implementation. It is expected that these clients will error if they receive some response that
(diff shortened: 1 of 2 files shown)


cb-heimdall:
<h1>🟡 Heimdall Review Status</h1>
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<summary>Denominator calculation</summary>
<table>
<tr>
<td>
<details>
<summary>Show calculation</summary>
<table>
<tr>
<td><code>1</code> if user is bot</td>
<td>0</td>
</tr>
<tr>
<td><code>1</code> if user is external</td>
<td>0</td>
</tr>
<tr>
<td><code>2</code> if repo is sensitive</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>.codeflow.yml</code>
</td>
<td>1</td>
</tr>
<tr>
<td>
Additional review requirements
<details>
<summary>Show calculation</summary>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td><code>1</code> if commit is unverified</td>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

